PR #2, opened 2015-08-13
Build process

msecret:
Ensures all dev dependencies are in package.json, which allows somebody to get everything up and running by running `npm install`. And moves all commands to npm run commands to ensure you don't have to find certain commands in `/node_modules/.bin` and for organization.


noahmanger:
Beautiful.

PR #6, opened 2015-08-15
First pass

noahmanger:
Contains all my work so far:

- Home page
- Browse pages
- Committee page

Everything is still a little rough around the edges, but getting the major pieces in place.


scss/components/_accordions.scss
```diff
@@ -0,0 +1,34 @@
...
+  clear: both;
+}
+
+.accordion__button {
+  position: absolute;
+  right: 1rem;
+  top: 1rem;
+}
+
+.accordion--collapsed {
```
msecret: This whole ruleset can be removed

noahmanger: Oh, I just forgot to add the styles. Adding...

scss/components/_charts.scss
```diff
@@ -0,0 +1,313 @@
+// the amount of padding at the bottom of vertical bar charts
+$chart-bottom: 30px;
+// the width of the axis for vertical bar charts
+$chart-axis-width: 120px;
+
+$tooltip-arrow-size: 8px;
+$tooltip-border-color: #999;
+
+.chart-container {
```
msecret: Why is there no `.chart` class? I feel like `.chart-container` could actually just be `.chart`.

scss/components/_charts.scss
```diff
@@ -0,0 +1,313 @@
...
+  @include align-items(flex-end);
+  padding-bottom: $chart-bottom;
+  padding-left: 0;
+  height: 200px;
+
+  @include media($med) {
+    padding-left: $chart-axis-width;
+  }
+
+  &.chart-series--short {
```
msecret: Does this ruleset have to be `&`? Can it just be on it's own?

scss/components/_charts.scss
```diff
@@ -0,0 +1,313 @@
...
+
+  .chart-series__axis__tick {
+    position: absolute;
+    text-align: right;
+    width: $chart-axis-width;
+    left: 0;
+    width: 100%;
+    height: 1em;
+
+    .chart-series__axis__tick__text {
```
msecret: I think this is too many levels of elements. Maybe something like this:

```
.chart-tick
.chart-tick__text
```

No need to fix now.

scss/components/_charts.scss
```diff
@@ -0,0 +1,313 @@
...
+      position: absolute;
+      left: -.5em;
+      right: 0;
+      bottom: 0;
+      border-top: 1px solid $neutral;
+    }
+  }
+}
+
+.chart-series__group__label {
```
msecret: Same as above, too many elements.

I also don't understand what "group" means in this context.

No need to fix now.

scss/components/_charts.scss
noahmanger: Ah yeah. I just copied this whole file entirely and only changed variable
names. It's probably due for a refactor but let's do that later. No need to
read too closely.

On Mon, Aug 17, 2015 at 3:16 PM, Marco Segreto <notifications@github.com>
wrote:

> In scss/components/_charts.scss
> <https://github.com/18F/fec-style/pull/6#discussion_r37243612>:
>
> > +
> > +    @include media($med) {
> > +      left: $chart-axis-width;
> > +      right: 20px;
> > +    }
> > +  }
> > +
> > +  .chart-series__axis__tick {
> > +    position: absolute;
> > +    text-align: right;
> > +    width: $chart-axis-width;
> > +    left: 0;
> > +    width: 100%;
> > +    height: 1em;
> > +
> > +    .chart-series__axis__tick__text {
>
> I think this is too many levels of elements. Maybe something like this:
>
> .chart-tick
> .chart-tick__text
>
> No need to fix now.
>
> —
> <https://github.com/18F/fec-style/pull/6/files#r37243612>.
>



-- 
Noah Manger

scss/components/_charts.scss
```diff
@@ -0,0 +1,313 @@
...
+  width: 100%;
+  padding-top: .25em;
+  text-align: center;
+
+  @include media($med) {
+    font-size: 1.2rem;
+  }
+}
+
+.chart-series__bar__tooltip {
```
msecret: I think `chart-tooltip` or `chart-series__tooltip` would suffice

No need to fix now.

scss/components/_dropdowns.scss
```diff
@@ -0,0 +1,118 @@
...
+//           <label class="dropdown__value" for="cycle-checkbox-year">2011 - 2012</label>
+//         </li>
+//       </ul>
+//     </div>
+//   </div>
+// </fieldset>
+// 
+// Styleguide components.dropdowns
+
+.button--dropdown {
```
msecret: I go back on forth with this but I feel like if this is in the dropdown file, it should be `dropdown--button`, `dropdown__button`.

noahmanger: Ah yeah, makes sense.

scss/components/_dropdowns.scss
```diff
@@ -0,0 +1,118 @@
...
+  position: absolute;
+  top: 3.2rem;
+  min-width: 100%;
+  background: $cream;
+  border: 2px solid $ivory-dark;
+  max-height: 30rem;
+  overflow: hidden;
+  z-index: 999;
+
+  li {
```
msecret: Is there any chance this won't always be an `<li>`?

scss/components/_dropdowns.scss
```diff
@@ -0,0 +1,118 @@
...
+    @include animation(fadeIn .2s ease-in);
+  }
+
+  input {
+    margin-left: 0;
+  }
+}
+
+.ps-scrollbar-y-rail {
```
msecret: What are these? Vendor styles?

scss/components/_hero.scss
```diff
@@ -0,0 +1,71 @@
...
+
+  @include media($med) {
+    @include span-columns(10);
+    @include shift(1);
+    padding-top: 6rem;
+
+    .search__radio {
+      @include span-columns(4 of 10);
+
+      .button--primary {
```
msecret: Does this have to be within `.search__radio`? It would help specificity if it was not.

scss/components/_dropdowns.scss
noahmanger: Yep :(

scss/components/_search-bar.scss
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    .combo__button {
+      height: 6rem;
+      width: 6rem;
+    }
+  }
+}
+
+// Typeahead styles
```
msecret: I've been trying to think of ways to deal with classes we can't control. One would be to have a `vendor-specific/typeahead file`. Just wanted to bring this up though, I don't think changes should be made now.

scss/components/_hero.scss
noahmanger: Fixed.

scss/mixins/_utilities.scss
```diff
@@ -0,0 +1,184 @@
...
+      color: $focus-color;
+    }
+  }
+  
+  select {
+    background-image: url('../img/i-down--' + $icon-color + '-contrast.png');
+    background-image: url('../img/i-down--' + $icon-color + '-contrast.svg');
+  }
+
+  input[type="checkbox"] + label::before,
```
msecret: It might be good to remove the `input` parts of these because this is in a mixin. So just `[type="checkbox"]`.

noahmanger: Done.

scss/modules/_entity-header.scss
```diff
@@ -0,0 +1,114 @@
...
+    padding: 1rem 0;
+  }
+}
+
+.entity__name {
+  font-weight: bold;
+}
+
+.entity__id {
+    color: #dde7f4;
```
msecret: Should this be using a color var?

scss/modules/_entity-header.scss
```diff
@@ -0,0 +1,114 @@
...
+}
+
+.entity__type {
+  line-height: 1.2;
+  display: block;
+  padding: .5rem 0;
+  float: left;
+
+  @include media($med) {
```
msecret: This rule isn't needed.

scss/modules/_filters.scss
```diff
@@ -0,0 +1,164 @@
...
+    width: 3rem;
+    height: 3rem;
+
+    &:hover, &:focus {
+
+    }
+  }
+}
+
+.filters__header {
```
msecret: Can this be removed?

scss/modules/_filters.scss
```diff
@@ -0,0 +1,164 @@
...
+  margin-top: 3rem;
+  padding-top: 2rem;
+  border-top: 2px solid $primary;
+
+  & + .filter {
+    padding-top: 0;
+  }
+}
+// Styles for when the filters are open
+.is-open {
```
msecret: I feel like if you see `is-open` you might apply that to things that aren't filters. Is that OK?

scss/modules/_filters.scss
noahmanger: I just left it because I still need to update these styles, but I'll add a TODO.

scss/modules/_filters.scss
noahmanger: I _think_ so, because it will only be matter if it's applied to the parent element?

noahmanger: Open to changing though.

scss/components/_dropdowns.scss
noahmanger: Good catch. Fixed.

noahmanger:
Does this need anything else or can we merge?

PR #15, opened 2015-08-19
Embedding fonts

noahmanger:
Embeds the webfonts. I used FontSquirrel's webfont generator with Base64 encoding a la these instructions: http://sosweetcreative.com/2613/font-face-and-base64-data-uri


msecret:
Glad to hear you used base64 encoding! At some point, we might want to improve performance with filament group's work  on this: https://www.filamentgroup.com/lab/font-events.html

noahmanger:
Great. Yeah I saw that. Seems like a good idea.

PR #17, opened 2015-08-19
Search select styling

noahmanger:
Styles for switching the search toggles to a select.


scss/components/_search-bar.scss
```diff
@@ -26,7 +26,7 @@
 
   @include media($med) {
     .combo__input {
-      width: calc(100% - 4.6rem);
+      width: calc(100% - 16.6rem);
```
msecret: What does 16.6 stand for? Seems like a random number. It is being set somewhere else?

noahmanger: It's the combined width of the select on the left and the button on the right.

PR #20, opened 2015-08-20
Restore value bar style.

jmcarp:
We lost the value bars inside the aggregate data tables with the shift to the new styles. This patch just restores the old styles--they don't look exactly right just now, but it's a start:
screen shot 2015-08-20 at 11 34 15 am
@noahmanger @jenniferthibault: thoughts on how these should look?

### DIFF
--- a/scss/modules/_datatables.scss
+++ b/scss/modules/_datatables.scss
@@ -17,7 +17,7 @@
     border-top: 0;
     margin: 0 1% 0 0;
     padding: 2rem 0;
-    width: 96%;    
+    width: 96%;
   }
 }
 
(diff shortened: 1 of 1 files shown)


scss/modules/_datatables.scss
```diff
@@ -92,7 +92,12 @@
     background-repeat: no-repeat;
     cursor: pointer;
     padding-left: 1rem;
-  }   
+  }
+
+  .value-bar {
+    height: 5px;
+    background-color: #205683;
```
noahmanger: Let's just use the variable `$primary` here.

noahmanger:
Oh thanks for catching this. Small note, but then i'll merge.

jmcarp:
Updated!

PR #25, opened 2015-08-20
Focus style

noahmanger:
Adds various :focus styles and a skip nav link


scss/modules/_site-header.scss
```diff
@@ -6,6 +6,24 @@
...
+.skip-nav {
+  @include transition(top, .2s);
+  background-color: $primary;
+  border: 2px solid $primary-contrast;
+  color: #fff;
+  font-family: $sans-serif;
+  left: 2rem;
+  padding: 1rem 2rem;
+  position: absolute;
+  top: -100px;
```
msecret: I'm guessing this needs to be a px value for a reason rather then rems?

noahmanger: Nope... I'll change it. Thanks.

PR #27, opened 2015-08-21
Build js for styleguide & refactored dropdowns

noahmanger:
1. Adds js to the static styleguide so you can demo interactive components
2. Refactors code for the filter dropdowns

Somewhat work in progress. I got the dropdown working as expected on the styleguide, and it worked at first on the app, but then inexplicably stopped working and now I'm going home :) 

Filing this in case someone wants to take a look. I tried to follow the pattern of the new glossary.js but this is newer stuff for me, so all feedback welcome.

### DIFF
--- a/js/dropdowns.js
+++ b/js/dropdowns.js
@@ -3,28 +3,27 @@
 /* global require, module, document */
 
 var $ = require('jquery');
-var perfectScrollbar = require('perfect-scrollbar/jquery') ($);
+var perfectScrollbar = require('perfect-scrollbar/jquery')($);
+
 var KEYCODE_ESC = 27;
 
 /**
  * Dropdown toggles
  * @constructor
- * @param {string} mainSelector - CSS selector for the fieldset that contains everything
+ * @param {string} selector - CSS selector for the fieldset that contains everything
  */
-
-function Dropdown(mainSelector) {
+function Dropdown(selector) {
   var self = this;
 
   self.isOpen = false;
-  self.hasPanel = true;
 
-  self.$body = $(mainSelector);
+  self.$body = $(selector);
   self.$selected = this.$body.find('.dropdown__selected');
   self.$button = this.$body.find('.dropdown__button');
   self.$panel = this.$body.find('.dropdown__panel');
 
   self.$button.on('click', this.toggle.bind(this));
-  self.$panel.on('change', 'input[type="checkbox"]', this.selectItem.bind(this));
+  self.$panel.on('change', 'input[type="checkbox"]', this.handleCheck.bind(this));
   $(document.body).on('click', this.handleClick.bind(this));
   $(document.body).on('keyup', this.handleKeyup.bind(this));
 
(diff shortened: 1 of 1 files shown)


noahmanger:
Update: this is working now.

fec-template/public/init.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+/* global require */
+
+var $ = require('jquery');
+var accordion = require('../../js/accordion.js');
```
jmcarp: By convention, leave off the `.js` or `.json` when `require`-ing dependencies.

js/dropdowns.js
```diff
@@ -0,0 +1,91 @@
...
+    self.hasPanel = true;
+
+    self.$body = $(mainSelector);
+    self.$selected = this.$body.find('.dropdown__selected');
+    self.$button = this.$body.find('.dropdown__button');
+    self.$panel = this.$body.find('.dropdown__panel');
+    self.$checkboxes = this.$panel.find('input[type="checkbox"]');
+
+    self.$button.on('click', this.toggle.bind(this));
+    self.$checkboxes.each(function(){
```
jmcarp: Instead of looping over checkboxes and binding a separate listener to each one, this might be a good use case for event delegation:
```javascript
self.$panel.on('change', 'input[type="checkbox"]', this.toggle.bind(this));
```
That way, we only have to use one listener, and checkboxes that move outside the panel no longer trigger the callback. Docs at http://learn.jquery.com/events/event-delegation/.

js/dropdowns.js
```diff
@@ -0,0 +1,91 @@
...
+    checkboxes[0].focus();
+    this.isOpen = true;
+}
+
+Dropdown.prototype.hide = function() {
+    this.$panel.attr('aria-hidden', 'true');
+    this.isOpen = false;
+}
+
+Dropdown.prototype.handleExit = function(ev) {
```
jmcarp: This callback handles two different event types, clicking a button and pressing escape. How about writing two smaller callbacks and binding the events separately? IMO this would improve readability and ease of testing.

js/dropdowns.js
```diff
@@ -0,0 +1,91 @@
...
+  ev.preventDefault();
+  var method = this.isOpen ? this.hide : this.show;
+  method.apply(this);
+}
+
+Dropdown.prototype.show = function() {
+    this.$panel.attr('aria-hidden', 'false');
+    this.$panel.perfectScrollbar({ 'suppressScrollX': true });
+    var checkboxes = this.$panel.find('input[type="checkbox"]');
+    checkboxes[0].focus();
```
jmcarp: Since we only care about the first checkbox here, what about something like:
```javascript
this.$panel.find('input[type="checkbox"]:first').focus();
```

js/dropdowns.js
```diff
@@ -0,0 +1,91 @@
...
+
+Dropdown.prototype.toggle = function(ev) {
+  ev.preventDefault();
+  var method = this.isOpen ? this.hide : this.show;
+  method.apply(this);
+}
+
+Dropdown.prototype.show = function() {
+    this.$panel.attr('aria-hidden', 'false');
+    this.$panel.perfectScrollbar({ 'suppressScrollX': true });
```
jmcarp: By convention, don't use quotes for object key names:
`this.$panel.perfectScrollbar({suppressScrollX: true});`

js/dropdowns.js
```diff
@@ -0,0 +1,91 @@
...
+
+Dropdown.prototype.hide = function() {
+    this.$panel.attr('aria-hidden', 'true');
+    this.isOpen = false;
+}
+
+Dropdown.prototype.handleExit = function(ev) {
+    var $target = $(ev.target);
+    if ( !this.$body.has($target).length ) {
+        this.hide.apply(this);
```
jmcarp: Why use `apply` here?

jmcarp:
This is looking good! I left some comments on the diff, all minor / stylistic. A few more general thoughts:
* Can we stick to two spaces for indentation for consistency? I've been using the airbnb style guide (more or less): https://github.com/airbnb/javascript#whitespace.
* What do you think about git-ignoring the compiled files? Not a big deal, but it's easy for the actual changes to get buried in the huge diffs we get with the compiled files checked in--this patch touched about 35,000 lines, but only a few hundred need to be reviewed.

noahmanger:
Great, thanks for the feedback. I'll take a look.

- Big +1 to two spaces. I'm not sure how this ended up with 4. Will fix.
- Also +1 to ignoring the compiled files.

noahmanger:
Merge conflicts solved and comments addressed.

PR #31, opened 2015-08-21
Build js

jmcarp:
This patch includes some very minor style changes--too small to waste time with individual comments on the original patch. @noahmanger, once this merged (if you agree with the changes), I think your patches to fec-style and the data app are good to go.
* Add missing semicolons
* Consistent variable naming and whitespace
* Removed a few unnecessary checks
* Removed focus next / previous on select--it made sense when we were showing checkboxes, but now I think it just looks weird.

### DIFF
--- a/js/dropdowns.js
+++ b/js/dropdowns.js
@@ -16,7 +16,6 @@ function Dropdown(selector) {
   var self = this;
 
   self.isOpen = false;
-  self.hasPanel = true;
 
   self.$body = $(selector);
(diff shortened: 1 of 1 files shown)


noahmanger:
Thanks for cleaning this up. The focus() stuff is important though as otherwise the page loses focus after selecting an item, which is bad for accessibility / ux.

jmcarp:
Focus restored!

PR #36, opened 2015-08-24
Percy integration

jmcarp:
* Update documentation
* Push to Percy on build success


jmcarp:
Looks like this one didn't get merged in with #33. @noahmanger or @msecret, can you take a look sometime? Not urgent, but it's a very small patch.

PR #37, opened 2015-08-24
Add a hover state for states.

jmcarp:
Note: For this to look good, we have to bring the hovered state path to
the top using javascript; else parts of the state border may be
obscured.

Question for @noahmanger or @onezerojeremy: should there be a hover state for state maps where the states aren't clickable? Under this patch, state borders are highlighted on hover whether clicking does a thing or not, but we can change easily if that would be clearer to users.

### DIFF
--- a/scss/components/_maps.scss
+++ b/scss/components/_maps.scss
@@ -3,6 +3,9 @@ path.shape {
   &.active {
     fill: $primary-contrast;
   }
+  &:hover {
+    stroke: $secondary-contrast;
+  }
 }
 
 .legend {


noahmanger:
For state maps with the tooltip, does that show up on click or hover? Seems like all of them will have some interaction on click, in which case having this change on hover makes sense.

onezerojeremy:
Ok so I'm thinking tooltip only on click as the tooltip will hide nearby states, but I think I'm struggling with the fact that outside of the map with table on the contributions by state, I don't recall where the map is (and thus what it should be doing). 

- for the map with table:
-- on click: bring table row of state name in focus. could also do tooltip, but probably unnecessary.
-- on hover: highlight borders, change cursor to finger, but maybe don't show tooltip as that's what the map is doing? 

- map without table 
-- on click: tooltip? 
-- on hover: highlight borders, change cursor to finger

jmcarp:
At the moment, the tooltip shows up on hover, but I'll change that to click if you prefer. @onezerojeremy: the map without state contributions is pretty buried at the moment! Here's how you find it:
* Go to an election page (e.g. https://fec-dev-web.18f.gov/elections/senate/ny/2012/)
* Scroll down to the "Compare By" header and click "Contribution Location"
* Click the "Map" toggle
* :boom:

noahmanger:
Sounds good to me. Let's try it out.

PR #40, opened 2015-08-25
Typeahead Filter

noahmanger:
This creates a new component for filter inputs that use the typeahead. Now, when selecting a committee from the typeahead, it appends a checkbox filter for that committee. 

see!

If there are any `committee_id`s on page load, it looks up the names of those committees and appends them. This makes it easy to filter by multiple committees.

Still needs tests.


noahmanger:
Waiting to implement this until later. Closing for now.

PR #42, opened 2015-08-25
Typeahead styles

jmcarp:
Miscellaneous cleanup:

* Bundle multiple API calls into one
* Store preset committees in hidden field to avoid changing the URL
* Update unit tests

### DIFF
--- a/tests/typeahead-filter.js
+++ b/tests/typeahead-filter.js
@@ -1,17 +1,23 @@
 'use strict';
 
-/* global require, document, describe, before, beforeEach, after, afterEach, it */
+/* global require, window, document, describe, before, beforeEach, after, afterEach, it */
+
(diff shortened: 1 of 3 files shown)


tests/typeahead-filter.js
```diff
@@ -1,17 +1,23 @@
 'use strict';
 
-/* global require, document, describe, before, beforeEach, after, afterEach, it */
+/* global require, window, document, describe, before, beforeEach, after, afterEach, it */
+
 var $ = require('jquery');
+var _ = require('underscore');
+
 window.$ = window.jQuery = $;
+_.extend(window, {
```
msecret: This feels like this should be in a `constants.js` or `globals.js` file. Is there a reason it isn't?

PR #46, opened 2015-08-27
Improving datatable and details panel

### LINKED ISSUE #24: Design loading state on tables
We should have some sort of different visual state on the tables if data is loading. An overlay, a spinner, something. Currently it just says loading above (which was meant as a placeholder).

noahmanger:
This patch does a few things:
- It ensures that if a table has few results, it still has a min-height to properly display the details pane [resolves https://github.com/18F/openFEC-web-app/issues/518] 
- Adds a fancy new loading animation [resolves #24]
- Fixes an issue caused by `display: table-cell` where after opening and then closing the details pane, the table would overflow its container.


jmcarp:
LGTM. Merging now, but @msecret, if you want to do a post-mortem review, comments welcome.

PR #53, opened 2015-09-01
Styles for mobile header and nav

noahmanger:
Here's a first pass at implementing the mobile nav based on these designs: https://github.com/18F/fec-alpha/issues/65

screen shot 2015-09-01 at 12 41 15 pm
screen shot 2015-09-01 at 12 41 42 pm
screen shot 2015-09-01 at 12 41 26 pm

It does not include the necessary ARIA-attributes for accessibility yet. I'll add a separate issue for it.

It also doesn't address the search / glossary issue on small screens, which I have some questions about. 

But if it looks good, I think we can merge and do those in future PRs.


scss/elements/_forms.scss
```diff
@@ -131,20 +131,20 @@ input[type="file"] {
 //
 // Styleguide elements.forms.checkboxes
 
-input[type="checkbox"],
-input[type="radio"] {
+[type="checkbox"],
```
msecret: Yay!

noahmanger: You were right! I ran into a super annoying thing with specificity and this solved it.

PR #62, opened 2015-09-03
Committee page styles

noahmanger:
Small style updates to committee pages and the components therein.

### DIFF
--- a/scss/components/_accordions.scss
+++ b/scss/components/_accordions.scss
@@ -24,9 +24,11 @@
 }
 
 .accordion__button {
+  margin-top: -1.5rem;
   position: absolute;
   right: 1rem;
-  top: 1rem;
+  top: 50%;
+
 }
 
 .accordion--collapsed {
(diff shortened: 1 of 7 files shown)


scss/components/_accordions.scss
```diff
@@ -24,9 +24,11 @@
 }
 
 .accordion__button {
+  margin-top: -1.5rem;
   position: absolute;
   right: 1rem;
-  top: 1rem;
+  top: 50%;
+
```
msecret: No space needed here.

scss/components/_charts.scss
```diff
@@ -274,9 +274,9 @@ $tooltip-border-color: #999;
 
 // Key
 .chart__header {
+  @extend .results-info;
```
msecret: Just to check, will `.chart__header` ever need to be extended?

noahmanger: I doubt it.

PR #65, opened 2015-09-04
Now it won't destroy the dropdown if there's a dropdown__subhead

### LINKED ISSUE #54: Dropdown is destroyed before all the options are gone
If a dropdown has a `dropdown__subtitle` separating list items, it gets destroyed even if there's more list items beyond.

noahmanger:
[resolves #54]

### DIFF
--- a/js/dropdowns.js
+++ b/js/dropdowns.js
@@ -27,7 +27,9 @@ function Dropdown(selector) {
   $(document.body).on('click', this.handleClickAndFocus.bind(this));
   $(document.body).on('keyup', this.handleKeyup.bind(this));
 
-  self.checkIfEmpty();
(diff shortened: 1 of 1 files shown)


js/dropdowns.js
```diff
@@ -95,8 +99,8 @@ Dropdown.prototype.removePanel = function() {
 };
 
 Dropdown.prototype.checkIfEmpty = function() {
-  if (this.$panel.find('.dropdown__item').length === 0) {
-    this.removePanel();
+  if (this.$panel.find('input').length === 0) {
```
jmcarp: This can be simplified to `return this.$panel.find('input').length === 0;`. I'll make this change on merging.

PR #67, opened 2015-09-04
Visual Style Tweaks

jenniferthibault:
Fumbling my way through small changes to the visual style here and there

So far:

 - Made large search bar text slightly smaller.The size difference between the drop-down text and input text is too large. Will try to find the drop-down text size next.
 - resizes large search box drop-down menu and button
 - adds space between bulleted list items for readability
 - left-aligns bullets with text box margins & grid
 - makes the photo attribution box slightly more transparent


jenniferthibault:
@noahmanger ready for review! I see there are merge conflicts, but this is a new part of GitHub to me

scss/modules/_datatables.scss
```diff
@@ -64,7 +64,8 @@
   td {
     @include transition(padding-left, .2s);
     @include u-truncate();
-    padding: 1rem;
+    padding: 1rem .5rem;
+    background: red;
```
noahmanger: I think you meant to leave this out :)

scss/components/_list-styles.scss
```diff
@@ -42,6 +43,11 @@
   }
 }
 
+// Alternate list styles
+ul.row {
```
noahmanger: Where is this used? We already have a `.row` class. If it's not the same, we should make a new class name for it.

noahmanger:
Made a couple comments. I can help you resolve conflicts soon.

scss/components/_list-styles.scss
jenniferthibault: It's on the beta home page around the bulleted list. I couldn't find class="row" in other files to edit, but maybe I missed it!

scss/modules/_datatables.scss
jenniferthibault: whoops, sure did!

PR #73, opened 2015-09-08
Styling the tooltips

noahmanger:
Styles the tooltips


scss/components/_tooltips.scss
```diff
@@ -0,0 +1,48 @@
...
+//
+// Markup:
+// <div class="tooltip">
+//  <div class="tooltip__title">Michigan</div>
+//  <div class="tooltip__value">$100,000</div>
+// </div>
+//
+// Styleguide components.tooltips
+
+.tooltip {
```
msecret: Should there be a max-width on them? They're appearing full width on the styleguide.

noahmanger: :+1:

noahmanger:
Ok, changes made.

jmcarp:
Because the tooltips are drawn in the middle of the corresponding state, the arrows seem to indicate the wrong state in some cases. For example, when you hover over Montana, the tooltip arrow points to Wyoming:
screen shot 2015-09-09 at 9 43 18 am

I can position the tooltips higher in the JS, or we can change the styles. What's your preference @noahmanger?

noahmanger:
Ah yeah. I'm trying to play around with it and it seems like the only solution that will really work is to do something like base it off the centroid of the state path. A quick google turned up this https://dillieodigital.wordpress.com/2013/02/13/quick-tip-getting-the-centroid-of-an-individual-svg-path-selection-in-d3/ .

noahmanger:
Can we merge these styles in and then sort out the positioning later?

jmcarp:
Yep, merging.

PR #79, opened 2015-09-10
Mobile nav aria attrs and focus

msecret:


js/site_nav.js
```diff
@@ -0,0 +1,36 @@
...
+  this.$body = $(selector);
+  this.$dropdowns = this.$body.find('ul ul');
+  this.$toggles = this.$body.find('.js-toggle');
+  this.assignAria();
+  this.$toggles.change(this.onChange);
+}
+
+SiteNav.prototype.assignAria = function() {
+  var $subLists = this.$body.find('ul ul'),
+      $links = this.$body.find('.js-nav-drop-link');
```
noahmanger: I don't see this applied in the web app

js/site_nav.js
```diff
@@ -0,0 +1,36 @@
...
+
+  this.$body.attr('aria-label', 'Site wide navigation');
+
+  $subLists.each(function() {
+    $(this).attr('aria-label', 'submenu');
+    $(this).attr('aria-hidden', true);
+  });
+
+  $links.attr('aria-haspopup', true);
```
noahmanger: Is 'has-popup' the right attribute if it's not actually a popup? I don't know.

noahmanger:
Might be a separate issue, but we need `:focus` styles for the menu button and the links inside. Probably not super important considering this will usually be a touch experience, but that'd be good to have anyway.

js/site_nav.js
jmcarp: Prefer one `var` declaration per variable: https://github.com/airbnb/javascript#13.2

tests/site_nav.js
```diff
@@ -0,0 +1,110 @@
+
+'use strict';
+
+/* global require, document, describe, before, beforeEach, after, afterEach, it */
+
+var chai = require('chai');
+var sinon = require('sinon');
+var sinonChai = require('sinon-chai');
+var expect = chai.expect;
+chai.use(sinonChai);
```
jmcarp: For clarity, prefer not to import modules we aren't using (sinon, sinonChai).

tests/site_nav.js
```diff
@@ -0,0 +1,110 @@
...
+        '</li>' +
+      '</ul>' +
+    '</nav>'
+    );
+    this.siteNav = new SiteNav('.js-site-nav');
+  });
+
+  describe('constructor()', function() {
+    it('should set body to jqueryized selector', function() {
+      expect(this.siteNav.$body).to.be.ok;
```
jmcarp: To be consistent with the description of this test, how about asserting that `$body` equals the jQuery element representing the selector?
```javascript
expect(this.siteNav.$body.is(this.$fixture('nav'))).to.be.true;
```

js/site_nav.js
```diff
@@ -0,0 +1,36 @@
...
+  this.$toggles.change(this.onChange);
+}
+
+SiteNav.prototype.assignAria = function() {
+  var $subLists = this.$body.find('ul ul'),
+      $links = this.$body.find('.js-nav-drop-link');
+
+  this.$body.attr('aria-label', 'Site wide navigation');
+
+  $subLists.each(function() {
```
jmcarp: Do we need to use `each` here? How about
```javascript
$sublists.attr({
  'aria-label': 'submenu',
  'aria-hidden': 'true'
});
```

js/site_nav.js
```diff
@@ -0,0 +1,36 @@
+'use strict';
+
+/* global require, module, document */
+
+function SiteNav(selector) {
+  this.$body = $(selector);
+  this.$dropdowns = this.$body.find('ul ul');
```
jmcarp: This variable is never used, but seems to be the same selector as `$subLists` in `assignAria`.

js/site_nav.js
```diff
@@ -0,0 +1,36 @@
...
+  this.$toggles = this.$body.find('.js-toggle');
+  this.assignAria();
+  this.$toggles.change(this.onChange);
```
jmcarp: `this.$toggles` is only used here, so may not need a variable assignment. We can also use event delegation to avoid binding a separate listener for each toggle:
```javascript
this.$body.on('change', '.js-toggle', this.onChange.bind(this));
```

fec-template/public/init.js
```diff
@@ -10,6 +10,7 @@ window.$ = window.jQuery = $;
 var accordion = require('../../js/accordion');
 var glossary = require('../../js/glossary');
 var dropdown = require('../../js/dropdowns');
+var siteNav = require('../../js/site_nav');
```
jmcarp: Prefer hyphenated file names, just for consistency.

js/site_nav.js
```diff
@@ -0,0 +1,36 @@
...
+
+  $subLists.each(function() {
+    $(this).attr('aria-label', 'submenu');
+    $(this).attr('aria-hidden', true);
+  });
+
+  $links.attr('aria-haspopup', true);
+}
+
+SiteNav.prototype.onChange = function(ev) {
```
jmcarp: This is a personal preference, but I like names like `handleChange` more than names like `onChange`. To me, `onChange` sounds like a method you might call to bind a change callback, but this method is the behavior we carry out to handle a change event. Your call, I'm just being obsessive.

js/site_nav.js
```diff
@@ -0,0 +1,36 @@
...
+  var $toggle = $(ev.currentTarget),
+      checked = $toggle.is(":checked"),
+      $list = $(ev.currentTarget).siblings('ul');
```
jmcarp: Same point as above re. one `var` per variable.

jmcarp:
Added some minor comments to the diffs. LGTM :pineapple:

js/site_nav.js
msecret: @noahmanger, its what I found in http://heydonworks.com/practical_aria_examples/#submenus

msecret:
@noahmanger agreed that we need focus styles for the nav. It was even hard to test whether focus was working without. I added a temporary rule. I think we can do it in another PR when the design is decided.

js/site_nav.js
noahmanger: Sweet. :+1:

PR #90, opened 2015-09-15
Restore dropdown and update unit tests.

jmcarp:
Please check out and verify that this is working as expected @noahmanger.

### DIFF
--- a/js/site-nav.js
+++ b/js/site-nav.js
@@ -2,11 +2,15 @@
 
 /* global require, module, document */
 
+var $ = require('jquery');
+
 function SiteNav(selector) {
   this.$body = $(selector);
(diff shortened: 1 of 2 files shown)


msecret:
This looks good to me, will wait for @noahmanger to check.

noahmanger:
Working well. We do need to make this work with tabbing though at some point. Could do it now or later.

jmcarp:
In the interest of demos working right away, we should probably merge. @noahmanger can you describe the desired tabbing behavior? I'll get to that next.

noahmanger:
When the parent `<li>` has focus, the submenu should be revealed, and then you should be able to tab through it.

PR #101, opened 2015-09-18
Unique tooltip

noahmanger:
Adds styles for tooltips underneath content (with the arrow above).

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -53,3 +53,4 @@ $z5: 500;
 $z-overlay: $z2;
 $z-navigation: $z5;
 $z-glossary: $z5;
+$z-tooltip: $z4;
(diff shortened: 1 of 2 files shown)


jmcarp:
One minor issue here: after checking the "unique only" checkbox, the associated tooltip stays on the screen until the user clicks somewhere else. If the user doesn't know they have to change focus, the filter element underneath can be obscured. I'm going to merge this now, but we might want to roll back the rule that shows tooltips on trigger focus @noahmanger.

noahmanger:
Good catch. We need it to be visible on focus somehow for keyboard accessibility. But maybe the way to do it is to have an icon and if you focus or hover on _that_ then it shows.

jmcarp:
SGTM! Do you have time to look at this soon, or should we file an issue for the backlog?

noahmanger:
Just added a note to this issue to make sure we address it: https://github.com/18F/openFEC-web-app/issues/689

PR #111, opened 2015-09-23
Adding simple table styles for cms pages

noahmanger:
Makes it so that basic tables in CMS content areas look like so:

screen shot 2015-09-23 at 11 45 58 am

@jenniferthibault any design feedback?

### DIFF
--- a/scss/components/_components.scss
+++ b/scss/components/_components.scss
@@ -33,5 +33,6 @@
 @import "search-results"; //@TODO: Document
 @import "search-controls"; //@TODO: Document
 @import "widgets"; //@TODO: Document
+@import "table-styles";  //@TODO: Document
 @import "richtext";
 @import "tooltips";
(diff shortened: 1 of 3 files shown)


jenniferthibault:
NIIIIICE!!!

@noahmanger I would like to tie table formatting in content modules to the table formatting in the data side for consistency. What that means in this instance, is changes the font family to Karla (bold for column categories, regular for data entries) and adding a closing line at the end of the last entry (under October 1–December 31). 

We also probably want to build in a title for the table, which might all look like this:

So it would look like this:

screen shot 2015-09-23 at 3 05 50 pm

What do you think?

noahmanger:
I like it! Update coming.

noahmanger:
Ok, fixed! 
screen shot 2015-09-23 at 1 18 36 pm

noahmanger:
Though, @emileighoutlaw note that we now want to include an `h3` before these tables.

PR #112, opened 2015-09-23
Adding other spending definitions

LindsayYoung:
Copied the definitions from the current FEC site. the urls are in the commit messages.  

First part of 18F/openFEC-web-app#573


jenniferthibault:
@emileighoutlaw I just asked @LindsayYoung and that `\` is what tells the javascript to show a quotation mark, not to interpret it as a command. 
> for JSON, the quotes are used as a way of structuring the data, so when you want to use a quote, you have to tell it that it really is a written quote and not the end of your object.
that is called escaping

That `\` is escaping a quote. Learn something new every day!

Another thing came up though, when she was looking at another part (Lindsay will have to point out which one): Do we want to use smart quotes here? Lindsay, I'd appreciate if you'd help fill in any context/details I've missed to help Emileigh out here

emileighoutlaw:
Oh Very cool! Thanks for enlightening me @LindsayYoung @jenniferthibault! Let me know what your thoughts are on smart quotes, Lindsay?

PR #113, opened 2015-09-23
Skip nav

noahmanger:
Implements a skip-nav link that looks for the first focus-able element in the `<main>` element on a page.

### DIFF
--- a/js/skip-nav.js
+++ b/js/skip-nav.js
@@ -20,3 +20,5 @@ Skipnav.prototype.focusOnTarget = function(e) {
     this.$target.focus();
   }
 };
+
+module.exports = {Skipnav: Skipnav};
--- a/js/skip-nav.js
+++ b/js/skip-nav.js
@@ -4,15 +4,22 @@
 
 var $ = require('jquery');
 
-function Skipnav(selector) {
-  var self = this;
-  self.$link = $(selector);
-  self.$target = self.findTarget();
-  $(document.body).on('click, keyup', self, this.focusOnTarget.bind(this));
+/**
+  * Skip nav link
+  * @constructor
+  * @param {string} anchor - CSS selector for the anchor element that will function as the skip nav
+  * @param {string} targetBody - CSS selector for the main content area to look for a focusable element in
+  */
+
+function Skipnav(anchor, targetBody) {
+  this.anchor = anchor;
+  this.$targetBody = $(targetBody);
+  this.$target = this.findTarget();
+  $(document.body).on('click, keyup', this.anchor, this.focusOnTarget.bind(this));
 };
 
 Skipnav.prototype.findTarget = function() {
-  return $('main').find('a, button, :input, [tabindex]').filter(':visible')[0];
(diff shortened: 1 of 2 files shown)


js/skip-nav.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+/* global require, module, document */
+
+var $ = require('jquery');
+
+function Skipnav(selector) {
+  var self = this;
+  self.$link = $(selector);
```
jmcarp: This variable doesn't seem to be used.

js/skip-nav.js
```diff
@@ -0,0 +1,24 @@
...
+  self.$target = self.findTarget();
+  $(document.body).on('click, keyup', self, this.focusOnTarget.bind(this));
+};
+
+Skipnav.prototype.findTarget = function() {
+  return $('main').find('a, button, :input, [tabindex]').filter(':visible')[0];
```
jmcarp: Prefer to minimize hard-coded selectors in generic UI components. How about moving "main", or even all of `$target`, to a parameter of the constructor?

js/skip-nav.js
```diff
@@ -0,0 +1,24 @@
...
+};
+
+Skipnav.prototype.focusOnTarget = function(e) {
+  if (e.keyCode === 13 || e.type === 'click') {
```
jmcarp: Does this focus on the target element every time the user clicks anywhere in the body?

js/skip-nav.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+/* global require, module, document */
+
+var $ = require('jquery');
+
+function Skipnav(selector) {
+  var self = this;
```
jmcarp: Probably no need to alias `self` to `this` here, since we aren't using any closures in this method.

js/skip-nav.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+/* global require, module, document */
+
+var $ = require('jquery');
+
+function Skipnav(selector) {
```
jmcarp: Could you add a brief description of what this is here? The name "Skipnav" is a little vague for me. A brief jsdoc-style comment would be great, a la https://github.com/18F/fec-style/blob/master/js/dropdowns.js#L11-L15.

js/skip-nav.js
noahmanger: Cool. Removed.

js/skip-nav.js
noahmanger: Done.

js/skip-nav.js
noahmanger: I was trying to do event delegation. I tweaked it so that the second parameter is the selector for the anchor element.

js/skip-nav.js
noahmanger: Done.

js/skip-nav.js
noahmanger: It is now, but renamed.

PR #123, opened 2015-09-26
Search bar styling

jenniferthibault:
This resolves https://github.com/18F/openFEC/issues/1222 by:
 - Making the drop-down field wider
 - Makes the search button slightly wider, and aligns it with the grid lines created by the glossary button above it.

Review:
@noahmanger 

I'm not sure if `.combo__select` and `.combo__input` needed to be adjusted for this, but it looked like them might need to be.

_______________________________

There is an additional change I'd like to make to the styling here, but I'm not sure how. I'd like to make the search text entry field less wide so that the whole search element is about 75% as wide as it is now. I think it is a bit too wide and becomes distracting because it doesn't align with other grid elements.

Here is current:
screen shot 2015-09-26 at 1 13 27 pm
Ideally the search bar would start on the left side right about where the last (Page TBD) begins, and ends where it currently ends as shown above. Like this:
screen shot 2015-09-26 at 1 21 14 pm

### DIFF
--- a/scss/components/_page-headers.scss
+++ b/scss/components/_page-headers.scss
@@ -49,7 +49,7 @@
 
 @include media($med) {
   .page-header {
-    padding: 1rem 4rem;
+    padding: 1rem 2rem;
 
     .search__container {
       display: block;
--- a/scss/components/_page-headers.scss
+++ b/scss/components/_page-headers.scss
@@ -56,7 +56,7 @@
     .search__container {
       display: block;
       float: right;
-      width: 50%;
+      width: 40rem;
     }  
   }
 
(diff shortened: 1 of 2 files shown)


noahmanger:
Nice. So it's not really possible (as far as I know) to get the width of it just so. The reason is because we're using a full-width flexible layout, so we can either set the bar take up a percentage of the total width of the window or have a fixed width that doesn't change. But because the nav items are "attached" to the left side and the search bar is "attached" to the right side, as the window size changes they're inevitably going to be off. Preference?

noahmanger:
Broken image url on candidate pages:

pasted_image_at_2015_09_26_02_00_pm

noahmanger:
Oops. Wrong PR for that comment. Merging.

PR #125, opened 2015-09-26
Universal style changes

jenniferthibault:
Small style tweaks overall including:
 - font sizes in committee tabs
 - reducing letter spacing in tabs
 - aligning check/radio buttons with text

### DIFF
--- a/scss/elements/_lists.scss
+++ b/scss/elements/_lists.scss
@@ -19,7 +19,9 @@
 ul,
 ol {
   list-style-type: none;
-  margin: 0;
+  margin-top: 0;
+  margin-bottom: 0;
+  margin-left: .25rem;
   padding: 0;
 }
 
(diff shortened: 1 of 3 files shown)


scss/elements/_lists.scss
```diff
@@ -19,7 +19,9 @@
 ul,
 ol {
   list-style-type: none;
-  margin: 0;
+  margin-top: 0;
+  margin-bottom: 0;
+  margin-left: .25rem;
```
noahmanger: Do you just want this to apply to bulletted lists? we often use `ul` for things that we don't want to look like lists, so if this is for bullet's it should go under `list--bulleted` in `list-styles.scss`.

jenniferthibault: Nope, I was looking at only radio buttons and check boxes here. Thanks for pointing this out though so I'll know where to look in the future!

jmcarp:
Is this ready for merge?

noahmanger:
I want to change the list margin stuff, but then yeah.

PR #128, opened 2015-09-26
Fixing contrast fail

### LINKED ISSUE #106: Minor contrast fail
screen shot 2015-09-22 at 2 05 28 pm

The #ffffff on #DD3900 is failing (4.48:1). Not a huge deal, but we should adjust.

noahmanger:
Fixes the contrast fail on these cards:

screen shot 2015-09-26 at 3 57 48 pm

Resolves #106

### DIFF
--- a/scss/components/_sidebar.scss
+++ b/scss/components/_sidebar.scss
@@ -86,7 +86,7 @@
   }
 
   .card {
-    @include u-font-color($ivory);
-    background-color: #DD3900;
+    @include u-font-color($cream);
+    background-color: #dd3700;
   }
 }


scss/components/_sidebar.scss
```diff
@@ -86,7 +86,7 @@
   }
 
   .card {
-    @include u-font-color($ivory);
-    background-color: #DD3900;
+    @include u-font-color($cream);
+    background-color: #dd3700;
```
jmcarp: This sounds like a good practice, but I'm going to merge now anyway.

PR #142, opened 2015-10-13
Fixing feedback position

noahmanger:
This sets the positioning of the feedback box based on viewport height, instead of percentages. This should ensure that it is always positioned below the bottom edge of the viewport when closed, and never above the edge of the top of the screen when open. 

Resolves https://github.com/18F/openFEC/issues/1269

Accidentally branched off of #135 instead of master, so let's merge that first.


.scss-lint.yml
```diff
@@ -90,10 +90,11 @@ linters:
     max_depth: 2
     severity: warning
   SelectorFormat:
-    enabled: true
-    convention: hyphenated_lowercase
+    enabled: false # strict_BEM doesn't seem to be supported by Hound
+    convention: strict_BEM
```
jmcarp: Does this work for you locally? Hound should just be calling `scss-lint`. If there's a bug, we should report it.

noahmanger: Yeah it does, but it's not super accurate I'm seeing (e.g. doesn't seem to like the `.block--modifier` format. So I'm just going to keep it off.

PR #143, opened 2015-10-14
Adds terms

emileighoutlaw:
Adds three terms: Receipt,
Disbursement, SSF, and
their definitions.


js/terms.json
```diff
@@ -87,4 +87,17 @@
     "glossary-term": "Communication cost",
     "glossary-definition": "52 U.S.C. 30118 allows \"communications by a corporation to its stockholders and executive or administrative personnel and their families or by a labor organization to its members and their families on any subject,\" including the express advocacy of the election or defeat of any federal candidate. The costs of such communications must be reported to the Federal Election Commission under certain circumstances."
    }
+   {
```
jmcarp: This is a JSON document, so list items need to be separated by commas. I'll make this change during merge, and I'll update the tests so that invalid JSON will cause failures in the future.

emileighoutlaw: Hey thanks.

PR #150, opened 2015-10-15
Adds styles for the staging banners

noahmanger:


scss/modules/_site-header.scss
```diff
@@ -1,7 +1,21 @@
...
 
+// Staging banner
+.banner {
+  background-color: $navy;
+  color: $cream;
+  font-family: $sans-serif;
+  padding: 1rem 0;
+
+  a {
+    color: $cream;
```
jmcarp: Is the idea that scss-lint wants (some) properties in alphabetical order? I wouldn't miss this warning if we lost it.

noahmanger: Yeah. I was trying to do that but I'm fine to lose it. It strikes me as really unimportant.

PR #156, opened 2015-10-21
Design review

jenniferthibault:
## Changes
 - adjusts spacing before and after table elements
 - makes page backgrounds white instead of almost white
 - changes card copy font on home page
 - aligns govt website flag icon
 - makes text in “official govt website” banner Karla instead of Gandhi

## Review
@noahmanger

### DIFF
--- a/scss/modules/_site-header.scss
+++ b/scss/modules/_site-header.scss
@@ -64,6 +64,7 @@
 .disclaimer {
   @include clearfix();
   border-bottom: 1px solid $primary;
+  font-family: $sans-serif;
   font-size: 1.2rem;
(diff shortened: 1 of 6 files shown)


scss/modules/_site-header.scss
```diff
@@ -84,6 +85,10 @@
   @include media($med) {
     display: inline-block;
   }
+
+  img {
```
noahmanger: Minor thing, but we've been putting the `@include media()` mixins at the bottom, so the `img { }

jenniferthibault: Oh great! I'll fix that

PR #157, opened 2015-10-22
Updating icons

jenniferthibault:
This updates some elements in the style guide, and changes the filings and registration icons.

@noahmanger eventually I’d like to add the example of the other Alerts to the style guide, but wasn’t quite sure how.

For now, I tried to make the size of the alert icon smaller through the image size, but that didn’t seem to have an effect on how it rendered here
screen shot 2015-10-22 at 4 30 15 pm


noahmanger:
Could you repost the screenshot?

jenniferthibault:
Whoops! Guess you can't upload screenshots through the desktop app. Fixed now.

noahmanger:
Ah. So that's set by the `background-size` property on `.message`. How big should it be? I can fix.

noahmanger:
Just double-checking: The new error and success icons look like this on the feedback widget. Is that what you want? 

image

image

noahmanger:
Also, how does this work for the message icon:
image

emileighoutlaw:
Sorry to slow the merge — made a couple small language tweaks here.

js/feedback.js
```diff
@@ -79,9 +79,9 @@ Feedback.prototype.submit = function(e) {
 
 Feedback.prototype.handleSuccess = function(response) {
   var message =
-    '<h2 class="feedback__title">Thank you for helping us improve betaFEC</h2>' +
+    '<h2 class="feedback__title">Thanks for helping us improve betaFEC</h2>' +
     '<p>This information has been reported on GitHub, where it is publicly visible. ' +
-    'You can track the status of your feedback <a target="_blank" href="' + response.html_url + '">here</a>.</p>';
+    '<a target="_blank" href="' + response.html_url + '">Track</a> the status of your feedback.</p>';
```
emileighoutlaw: @noahmanger, why isn't it working to make "Track" the link? Did I mess up the HTML?

jenniferthibault:
Thanks for catching this @noahmanger ! I didn't anticipate the naming implications for these very well, and that's NOT what I was trying to do on the feedback module. The bright side is that this identified a pattern we weren't aware of in alerts. The downside is that it means it's a little more complicated. The bright side again is that I have ideas! Fleshing them out over in #158 for discussion, then we can figure out what the best way to get them into a PR is

PR #165, opened 2015-10-31
Use karma

jmcarp:
Use karma test runner.

Karma is a better-supported and fuller-featured test runner than
mochify. It's also not tied to browserify, which would smooth the
transition to another build tool should we go in that direction.


jmcarp:
@noahmanger this is ready for review. Maybe @msecret can help review--IIRC, you recommended karma at some point, so this should be up your alley.

noahmanger:
It works for me but @msecret should definitely review. I wouldn't mind getting walked through it at some point, too, so I can better understand what's going on.

jmcarp:
SGTM. By the way, the karma changes here are basically copied in https://github.com/18F/openFEC-web-app/pull/900, which @shawnbot has graciously agreed to review. So if that patch gets merged first, we can probably call this one good.

@msecret: how's your availability lately?

msecret:
This looks good to me.

My availability is a little sparse these days, maybe I'll discuss with you guys in the chatroom.

jmcarp:
@noahmanger: re-running tests on code changes and the in-browser debugger are both pretty sweet. We can spend a few minutes going over usage sometime if you want. Might be interesting to @LindsayYoung too.

noahmanger:
I'm going to merge and then you can walk us through later.

PR #167, opened 2015-11-03
Set focus on what triggered glossary

noahmanger:
This finds whatever element opened the glossary and then when the
glossary closes it restores focus to whatever triggered it.

Resolves https://github.com/18F/FEC/issues/102

### DIFF
--- a/js/glossary.js
+++ b/js/glossary.js
@@ -49,9 +49,9 @@ function Glossary(terms, selectors) {
   self.populate();
   self.linkTerms();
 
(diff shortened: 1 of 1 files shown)


noahmanger:
Hmmm. `npm test` is failing on Travis and failing locally for me. Any idea what's going on?

jmcarp:
Will take a look tonight or tomorrow morning. FWIW, I think we'll get cleaner test output with #165.

PR #168, opened 2015-11-03
Fix and expand glossary tests.

jmcarp:
* Pass mock events to glossary methods as needed
* Test that hiding the glossary focuses the trigger
* Short-circuit focus if trigger is not set

cc @noahmanger

### DIFF
--- a/js/glossary.js
+++ b/js/glossary.js
@@ -41,6 +41,7 @@ function Glossary(terms, selectors) {
   self.$body = $(self.selectors.body);
   self.$toggle = $(self.selectors.toggle);
   self.$search = this.$body.find('.glossary__search');
+  self.$trigger = null;
 
   // Initialize state
   self.isOpen = false;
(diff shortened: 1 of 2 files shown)


js/glossary.js
```diff
@@ -127,13 +128,13 @@ Glossary.prototype.show = function(e) {
 Glossary.prototype.hide = function() {
   this.$body.removeClass('is-open').attr('aria-hidden', 'true');
   this.$toggle.removeClass('active');
-  this.$trigger.focus();
+  this.$trigger && this.$trigger.focus();
```
noahmanger: What's the point of this change?

msecret: So only trigger focus if the trigger exists. I like this pattern, maybe we should relax the houndci warnings?

msecret:
LG

js/glossary.js
jmcarp: Right. I actually got rid of that warning on the webapp--I can copy over our linter config here to avoid the complaint.

PR #170, opened 2015-11-04
Refactored nav

noahmanger:
This cleans up the nav code and ensures that it's keyboard accessible. 
- Replaces the complicated hidden checkbox method for showing collapsible sublists on the mobile menu with a simple button trigger
- Refactors the javascript to include `showSublist()` and `hideSublist()` functions that can be called by either clicking the `.js-sublist-toggle` button on mobile, hovering with a mouse, or focusing on element with the keyboard.
- Refactors the Sass to clean it up and group all media queries together, which is the new pattern we're following
- Updates tests

Resolves https://github.com/18F/openFEC-web-app/issues/904


tests/site_nav.js
```diff
@@ -92,33 +88,44 @@ describe('SiteNav', function() {
...
-      var $list = this.$fixture.find('#site-menu ul');
-      var testEv = {currentTarget: $toggle};
-
-      this.siteNav.handleChange(testEv);
-      expect($list.attr('aria-hidden')).to.not.equal($toggle.is(':checked'));
+  describe('toggle()', function() {
+    it('should show and hide the menu', function() {
+      var $nav = this.$fixture.find('.js-site-nav');
+      this.siteNav.toggle();
+      expect($nav.attr('class')).to.include('is-open');
```
jmcarp: Prefer something like `expect($nav.hasClass('is-open')).to.be.true`. That's a little more robust that checking substring membership, since this logic would also match a hypothetical class like "not-is-open-lol".

js/site-nav.js
```diff
@@ -47,21 +49,33 @@ SiteNav.prototype.hide = function() {
...
 };
 
-SiteNav.prototype.handleChange = function(e) {
-  var $subToggle = $(e.currentTarget);
-  var checked = $subToggle.is(':checked');
-  var $list = $(e.currentTarget).siblings('ul');
+SiteNav.prototype.toggleSublist = function(e) {
+  var method = this.$openSublist ? this.hideSublist : this.showSublist;
+  var $target = $(e.target);
+  var $sublistParent = $target.hasClass('js-sublist-parent') ? $target : $target.closest('.js-sublist-parent');
```
jmcarp: Since `closest` tests the target element as well as its parents, we should be able to simplify this as `var $sublistParent = $target.closest('.js-sublist-parent')`:

https://api.jquery.com/closest/

scss/modules/_nav.scss
```diff
@@ -297,92 +241,134 @@
...
+        color: $ivory;
+      }
     }
 
-    & > .nav-toggle__input,
-    & > .nav-toggle__label {
-      display: none;
+    &.is-disabled {
+      .site-nav__link {
+        color: #737373;
```
jmcarp: Seems like a good suggestion.

tests/site_nav.js
```diff
@@ -79,8 +75,8 @@ describe('SiteNav', function() {
       expect(this.$subLists.last().attr('aria-hidden')).to.equal('true');
     });
 
-    it('should assign aria-haspopup to all links to the sub lists', function() {
-      var $link = this.$fixture.find('#site-menu .js-nav-drop-link');
+    it('should assign aria-haspopup to all toggle buttons for the sub lists', function() {
+      var $link = this.$fixture.find('.js-sublist-toggle');
```
jmcarp: This test doesn't seem to do exactly what the description says--we're asserting that there's at least one `.js-sublist-toggle`, and that the first one has the "aria-haspopup" attribute. Which is fine if there's exactly one toggle button. But if not, we might use something like the following:

```javascript
var $toggles = this.$fixture.find('.js-sublist-toggle');
var $popupToggles = this.$fixture.find('.js-sublist-toggle[aria-haspopup="true"]');
expect($toggles.length).to.be.ok;
expect($toggles.length).to.equal($popupToggles.length);
```

tests/site_nav.js
```diff
@@ -92,33 +88,44 @@ describe('SiteNav', function() {
...
-    beforeEach(function() {
-      this.$withDropdown = this.$fixture.find('.site-nav__item--with-dropdown');
-      this.$list = this.$fixture.find('#site-menu ul');
+  describe('toggleSublist()', function() {
+    it('should show and hide the sublist', function() {
+      var $sublistParent = this.$fixture.find('.js-sublist-parent');
+      var $sublistToggle = this.$fixture.find('.js-sublist-toggle');
+      var testEv = {target: $sublistToggle};
+      this.siteNav.toggleSublist(testEv);
+      expect($sublistParent.find('ul').attr('aria-hidden')).to.equal('false');
```
jmcarp: Since `hideSublist` and `showSublist` each do a few different things, it might be worth adding helpers like `isOpen` and `isClosed` that check all the relevant DOM settings, plus the `isOpen` variable. See https://github.com/18F/fec-style/blob/master/tests/dropdowns.js#L15-L23 for examples.

js/site-nav.js
```diff
@@ -47,21 +49,33 @@ SiteNav.prototype.hide = function() {
...
+  $sublistParent.removeClass('is-open');
+  $sublistParent.find('ul').attr('aria-hidden', true);
+  this.$openSublist = null;
 };
 
-SiteNav.prototype.handleMouseExit = function(e) {
-  $(e.currentTarget).find('.js-toggle').prop('checked', false).change();
+SiteNav.prototype.handleFocus = function(e) {
+  var $target = $(e.target);
+  var $sublistParent = $target.closest('.js-sublist-parent') || false;
```
jmcarp: The `|| false` part probably isn't necessary here. We only get to that branch if the value of `closest(...)` is false-y, and in that case, we already have a false-y value to be checked on line 76.

js/site-nav.js
```diff
@@ -7,18 +7,20 @@ var $ = require('jquery');
...
+  this.$openSublist = null;
 
   this.assignAria();
 
   this.$toggle.on('click', this.toggle.bind(this));
-  this.$body.on('change', '.js-toggle', this.handleChange.bind(this));
-  this.$body.on('mouseenter', 'li:has(.js-toggle)', this.handleMouseEnter.bind(this));
-  this.$body.on('mouseleave', 'li:has(.js-toggle)', this.handleMouseExit.bind(this));
+  this.$body.on('click', '.js-sublist-toggle', this.toggleSublist.bind(this));
+  this.$body.on('mouseenter', '.js-sublist-parent', this.toggleSublist.bind(this));
```
jmcarp: For clarity, how about binding "mouseenter" to `showSublist` and "mouseleave" to `hideSublist`?

js/site-nav.js
```diff
@@ -7,18 +7,20 @@ var $ = require('jquery');
...
+  this.$body.on('mouseleave', '.js-sublist-parent', this.toggleSublist.bind(this));
+  $(document.body).on('focusin', this.handleFocus.bind(this));
 }
 
 SiteNav.prototype.assignAria = function() {
   var $subLists = this.$body.find('ul ul');
-  var $links = this.$body.find('.js-nav-drop-link');
+  var $links = this.$body.find('.js-sublist-toggle');
```
jmcarp: This variable name is a little confusing (to me, not being familiar with the code). How about either renaming, or skipping the assignment entirely:

```javascript
this.body.find('.js-sublist-toggle').attr(...);
```

js/site-nav.js
```diff
@@ -7,18 +7,20 @@ var $ = require('jquery');
...
   this.assignAria();
 
   this.$toggle.on('click', this.toggle.bind(this));
-  this.$body.on('change', '.js-toggle', this.handleChange.bind(this));
-  this.$body.on('mouseenter', 'li:has(.js-toggle)', this.handleMouseEnter.bind(this));
-  this.$body.on('mouseleave', 'li:has(.js-toggle)', this.handleMouseExit.bind(this));
+  this.$body.on('click', '.js-sublist-toggle', this.toggleSublist.bind(this));
+  this.$body.on('mouseenter', '.js-sublist-parent', this.toggleSublist.bind(this));
+  this.$body.on('mouseleave', '.js-sublist-parent', this.toggleSublist.bind(this));
+  $(document.body).on('focusin', this.handleFocus.bind(this));
```
jmcarp: For clarity, how about renaming `handleFocus` to something that better captures what's going on, like `handleFocusBody`?

js/site-nav.js
```diff
@@ -7,18 +7,20 @@ var $ = require('jquery');
...
   this.$toggle = this.$body.find('.js-nav-toggle');
+  this.$openSublist = null;
 
   this.assignAria();
 
   this.$toggle.on('click', this.toggle.bind(this));
-  this.$body.on('change', '.js-toggle', this.handleChange.bind(this));
-  this.$body.on('mouseenter', 'li:has(.js-toggle)', this.handleMouseEnter.bind(this));
-  this.$body.on('mouseleave', 'li:has(.js-toggle)', this.handleMouseExit.bind(this));
+  this.$body.on('click', '.js-sublist-toggle', this.toggleSublist.bind(this));
```
jmcarp: I was trying to understand what was going on here better, and I realized that this class doesn't (yet) appear in the webapp or cms. Are we planning to use it? Is it in the examples in this repo?

noahmanger: Oops, yes. It's in a branch I was working on for the web-app but haven't pushed yet.

jmcarp:
LGTM; left comments on the diffs. At a higher level, I think we could make this easier to understand with slightly more descriptive names and docstrings. For example, clarifying that this is for smaller screens and is distinct from the desktop nav would be helpful (I forget sometimes)--what about naming this `MobileNav`, `SiteNavMobile`, etc.? We could also include a brief description of what the widget is for in a docstring on the constructor.

noahmanger:
Thanks for the thorough review. I'm still working on edits in between meetings. Will add more description too.

tests/site_nav.js
noahmanger: Replaced this with helpers to check all properties.

tests/site_nav.js
noahmanger: :+1:

tests/site_nav.js
noahmanger: Makes sense. Testing philosophy question though: so in this case, there is only one toggle, but in the real world there may be more. Do you write tests for what you could encounter or for how you've set up the test DOM? (Does that even make sense?)

js/site-nav.js
```diff
@@ -4,35 +4,44 @@
 
 var $ = require('jquery');
 
+/** SiteNav module
```
jmcarp: Extremely minor, but jsdoc-style comments do spacing just a little differently--we'd want to dedent lines 8-12 by one space. If you're using sublime, check out https://github.com/spadgos/sublime-jsdocs or something similar.

js/site-nav.js
```diff
@@ -4,35 +4,44 @@
...
+  */
+
 function SiteNav(selector) {
   this.$body = $(selector);
   this.$toggle = this.$body.find('.js-nav-toggle');
+  this.$openSublist = null;
 
   this.assignAria();
 
+  /** Open and close the menu on mobile **/
```
jmcarp: The `/** ... */` comment style is for jsdoc-formatted comments specifically. For unstructured comments like this, I think `// ...` or `/* ... */` works fine.

jmcarp:
The code looks good to me, but I noticed an issue when I checked this out with https://github.com/18F/openFEC-web-app/pull/925: the side nav hides the hamburger icon when it's expanded, so that you can't hide the menu once it's open:

screen shot 2015-11-05 at 2 45 10 pm

@noahmanger are you seeing the same behavior?

noahmanger:
That's just an issue with the dev banner. The menu is positioned absolutely from the top, but because the banner pushes everything down, it's covering the menu. If you delete that banner it should work correctly.

PR #177, opened 2015-11-07
Adding classes for inline icons

noahmanger:
Small class to center the inline icons used in the new feedback links in the footer.

image

@jenniferthibault


jenniferthibault:
Thanks Noah!

PR #179, opened 2015-11-11
Download buttons

noahmanger:
Adds a very rudimentary modal js component, template and styles to be used for download button testing. 

Eventually, I'm thinking we'll expand on this component so it can work for a variety of different modals, but this will get the job done for now.


noahmanger:
Closing this in favor of openFEC-web-app using this specific branch for the test.

Issue #173, opened 2015-11-05
To help users leave better feedback: consider language edits to feedback tool


emileighoutlaw:
#### Current content
<img width="541" alt="screen shot 2015-11-05 at 11 17 10 am" src="https://cloud.githubusercontent.com/assets/10552702/10974187/d410e3b4-83ae-11e5-8958-f8bfbc6a832a.png">

#### Tasks to completion:
 - [ ] Draft preloaded text (if we want)
 - [ ] Propose edits to feedback prompts
 - [ ] Finalize all content   
 - [ ] Implement @emileighoutlaw

jenniferthibault:
I would love to create more clear distinction between the first and second questions (What were you doing?/What happened?) Seems like now these are a bit redundant, and the pattern of feedback we've been receiving seems to show that users aren't really separating the fields in a meaningful way.

What if....something along these lines?

**Tell us about what you were trying to do**
*I am a ______ interested in ______. I was trying to....*

**What could be improved or corrected about that task?**

**Or, just give us some feedback** or **General feedback**


OR!


**Tell us about yourself**
*I am a ______ interested in ______.* 

**Tell us about what you were trying to do on the site, and how it could be improved**

**Or, just give us some feedback** or **General feedback**

emileighoutlaw:
Great draft, Jen! Here's my thinking:

I like adding a note at the top that every field is optional. From there, it seems there are three pieces of info we'd like.

1. Who the person is
2. What they were doing/what went wrong
3. General feedback

For 2. to be complete we need to know both what they were doing and what went wrong. So it seems wise to keep them as one field. People either fill it out, or they leave general feedback.

As a jumping off point:

1.
>**Help us improve betaFEC**
>*All fields are optional. Don't include sensitive information...*

>Tell us about yourself
>*I'm a ______ interested in ________.*

> What were you trying to do, and how can we improve it?

> General feedback

2.
>**Help us improve betaFEC**
>*All fields are optional. Don't include sensitive information...*

>Tell us about yourself
>*I'm a ______ interested in ________.*

> How can we improve your experience?
> *I was trying do ___________ when _________ happened.*

> General feedback



We could also make the "Tell us about yourself" the last field, since it's a nice thing to have but not critical.

noahmanger:
I like this idea! I'm fine with either construction, and I think I like the idea of having the "I'm a ___" question last.

[...]

PR #182, opened 2015-11-14
Updates feedback language

noahmanger:
- And adjusts styles to accommodate longer fields
- Adds a couple extra classes

Resolves https://github.com/18F/fec-style/issues/173

Do not merge until web-app and CMS have been updated.


noahmanger:
Just doing a sanity check: we'll need to make sure we make the corresponding web app changes before deploying to production again, right (as in the case of a hotfix)? Otherwise it'll break the tool? Another reason to get versioning working with this.

jmcarp:
Right. To start doing versioning, all we need to do is update the version number in `package.json` according to http://semver.org/ when we push to master, and start pinning to specific version numbers in the data app.

PR #186, opened 2015-11-17
Version using semver.

jmcarp:
Proposal: Version fec-style using semantic versioning. Bug fixes
increment the patch, non-breaking changes increment the minor version, and
breaking changes increment the major version. To support flexible
version ranges (e.g. ~1.0.0), fec-style should be published to npm. If
we go in this direction, everyone who needs to be able to push changes
will need an npm account with permissions to update the fec-style
package. I'm arbitrarily calling this version 1.0.0.

Thoughts @noahmanger @shawnbot @msecret? I'm also hoping that versioning upstream dependencies gets simpler as we break more components out of fec-style and into their own packages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ When merging into master, update the package version and push to npm:
     npm publish
     git push --tags
 
+Use `npm version minor` or `npm version major` for minor and major updates, respectively. For details on npm versioning, see `npm version --help`.
+
 Downstream applications should pin versions as appropriate. For example, to get bug fixes but not new features, pin to the minor version:
 
     {
--- a/README.md
+++ b/README.md
@@ -3,13 +3,13 @@
 
 
 ## Campaign finance for everyone
-The Federal Election Commission (FEC) releases information to the public about money that’s raised and spent in federal elections — that’s elections for US president, Senate, and House of Representatives. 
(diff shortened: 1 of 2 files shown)


noahmanger:
Nice. I'm into this. The only question I have now is how to categorize Sass changes? I might need to see some examples around what counts as a breaking change vs. non-breaking, considering the possibility of unintended side effects.

jmcarp:
Some breaking changes will be obvious--if we change classes or DOM structure such that consumers will have to change their markup, we definitely have breaking changes. But yeah, I can see how it could be hard to tell whether a style change here would break things downstream. In practice, I think we'll want to continue manually testing all changes to fec-style against a local version of openFEC-web-app, which would hopefully help us catch unexpected side effects.

This may also be a good issue to bring up in #g-frontend.

noahmanger:
Can you resolve the conflicts? And then I'll merege.

jmcarp:
Conflicts resolved.

Related to this: to make it safe to merge breaking changes into master, I think we should hotfix the data site and CMS to pin to the current commit; then merge in pending (potentially breaking) changes; then release to npm. Then we'll install from npm and pin to the appropriate (fuzzy) version going forward. Thoughts?

noahmanger:
That sounds good with the one exception that we should do a thorough review to make sure we haven't accidentally introduced breaking changes since production was last deployed.

PR #189, opened 2015-11-19
Styling new cycle-selects

noahmanger:
Here's the styles for the new cycle selects, based off https://github.com/18F/openFEC-web-app/issues/942

Some of the alignment isn't quite right (for various reasons I can explain), but it's getting close. I'd vote to merge now and make fine-tuning changes after we've verified that in general this is working for folks.

gif

Requires https://github.com/18F/openFEC-web-app/pull/950

### DIFF
--- a/scss/components/_cycle-select.scss
+++ b/scss/components/_cycle-select.scss
@@ -1,19 +1,59 @@
 // Cycle Select
 //
-// Styling for when it's inline with a heading
-//
-// Markup:
-// <h2>Section title
-//   <select class="cycle-select js-cycle" aria-label="Two-Year Period" name="cycle">
-//     <option>2012 - 2013</option>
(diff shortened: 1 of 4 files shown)


scss/components/_cycle-select.scss
```diff
@@ -1,19 +1,59 @@
...
-// </h2>
-//
 // Styleguide components.cycle-select
 
 .cycle-select {
-  display: inline-block;
-  margin-bottom: 1rem;
-}
\ No newline at end of file
+  @include span-columns(6);
```
jmcarp: @noahmanger: if we merge this now and happen to redeploy production, will the cycle select styling be off?

noahmanger: Yep. This is a breaking change.

PR #191, opened 2015-11-24
Styling multicycles

noahmanger:
Styles for the candidate cycle selects.


jmcarp:
Looks good. Merging into develop and not master.

PR #202, opened 2015-12-09
Candidate styles

noahmanger:
Small style changes to implement https://github.com/18F/openFEC-web-app/issues/976

Goes with https://github.com/jmcarp/openFEC-web-app/pull/17

### DIFF
--- a/scss/components/_callouts.scss
+++ b/scss/components/_callouts.scss
@@ -1,5 +1,5 @@
 // Callout
-// 
+//
 // Action-oriented components
 //
 // Styleguide components.callouts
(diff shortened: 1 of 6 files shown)


jmcarp:
Going to merge for now, but it seems like anything as specific as `candidate_page` belongs in `openfec-web-app` rather than the shared styles. To my mind, this repo holds *shared* styles between apps, not necessarily all styles. Anyway, let's discuss later.

noahmanger:
That's a good point! I had just been thinking of it as all styles. Good to consider. TBH it's still a little bit of a hack. I'd like for it to be more of a generic component, but while we're still working out kinks in the design it's easier to have it be specific for now and then abstract it out.

PR #209, opened 2015-12-16
Typo fix

emileighoutlaw:
Eliminates stray space.

[ci skip]

cc @jenniferthibault ?


js/terms.json
```diff
@@ -61,7 +61,7 @@
...
+    "glossary-definition": "An entity that meets one of the following conditions:<ul class='list--bulleted'><li>An authorized committee of a candidate (see definition of candidate)</li><li>A state party committee or nonparty committee, club, association or other group of persons that receives contributions or makes expenditures, either of which aggregate over $1,000 during a calendar year</li><li>A local unit of a political party (except a state party committee) that: (1) receives contributions aggregating over $5,000 during a calendar year; (2) makes contributions or expenditures either of which aggregate over $1,000 during a calendar year; or (3) makes payments aggregating over $5,000 during a calendar year for certain activities which are exempt from the definitions of contribution and expenditure (11 CFR 100.80, 100.87 and 100.89; 11 CFR 100.140, 100.147 and 100.149)</li><li>Any separate segregated fund upon its establishment. 11 CFR 100.5.</li><ul>"
```
jmcarp: No reason not to fix this, but as a pedantic aside, multiple spaces in HTML are rendered as a single space. Which is a neat way to avoid the issue of how many spaces go after a sentence (correct answer: one). And yes, this isn't an HTML file, but the content here is destined to be rendered as HTML later on.

All of which is to say: :+1:

jenniferthibault: I didn't know that, so cool!

PR #213, opened 2015-12-23
First pass at calendar styles

noahmanger:
Work in progress. Don't merge.


js/dropdowns.js
```diff
@@ -8,32 +8,41 @@ var perfectScrollbar = require('perfect-scrollbar/jquery')($);
...
+  self.opts = $.extend({}, defaultOpts, opts);
 
   self.isOpen = false;
 
   self.$body = $(selector);
-  self.$selected = this.$body.find('.dropdown__selected');
   self.$button = this.$body.find('.dropdown__button');
   self.$panel = this.$body.find('.dropdown__panel');
 
+  if (self.opts.checkboxes === true) {
```
jmcarp: Could be simplified to `if (self.opts.checkboxes)`

js/filter-tags.js
```diff
@@ -0,0 +1,46 @@
...
+    '</button>' +
+  '</li>',
+  {interpolate: /\{\{(.+?)\}\}/g}
+);
+
+function Tag($input, filter) {
+  this.$input = $input;
+  this.key = $input.attr('id');
+  this.filter = filter;
+  this.checkOrRadio = $input.attr('type') === 'checkbox' || 'radio';
```
jmcarp: This might not be doing what you want it to. If this variable is meant to capture whether the input is a checkbox or radio button (true) or not (false), then we might want something like `['checkbox', 'radio'].indexOf($input.attr('type')) !== -1`.

js/filter-tags.js
```diff
@@ -0,0 +1,46 @@
...
+  this.text = this.getText($input);
+
+  this.$content = $(TAGTEMPLATE({text: this.text}));
+
+  this.$content.on('click', 'button', this.remove.bind(this));
+};
+
+Tag.prototype.getText = function($input) {
+  var text;
```
jmcarp: No need for the `text` variable here.

js/filter-panel.js
```diff
@@ -5,9 +5,11 @@ var _ = require('underscore');
 
 var FilterSet = require('./filter-set').FilterSet;
 var accessibility = require('./accessibility');
+var filterTags = require('./filter-tags');
```
jmcarp: Good catch, Hound! Not sure why we got three of these messages. Anyway, we should probably make this change and verify that any lingering unused code is gone @noahmanger.

noahmanger: :+1:

js/filter-panel.js
```diff
@@ -25,21 +27,24 @@ function FilterPanel(options) {
...
-  if ($('body').width() > 768) {
+  if ($(document).width() > 768) {
     this.show();
   } else if (!this.isOpen) {
     this.hide();
   }
 };
 
+FilterPanel.prototype.setHeight = function() {
+  if ( $(document).width() > 768 && this.$dataContainer.height() > this.$body.height() ) {
```
jmcarp: Could this 768 be a named constant? Do we use it anywhere else?

jmcarp:
Assigning to @ccostino for review. This patch includes lots of stuff (probably too much), so let's focus review on changes to filter behavior. I think it will be quicker to walk through the proposed changes over screenshare than to write everything up on GH. @ccostino, would you mind booking 15 minutes on my calendar sometime to talk this through?

js/filter-panel.js
noahmanger: Oh yep, it could be. We don't use it much, but every once in a while. Should we move the `GLOBALS` object to helpers.js in fec-style?

noahmanger: Or we could have a new object there for `BREAKPOINTS`

ccostino:
@jmcarp Yes, will do.  I will try to find a time for both of us in the next couple of days!

js/filter-panel.js
```diff
@@ -25,21 +28,25 @@ function FilterPanel(options) {
...
-  if (!_.isEmpty(this.filterSet.serialize())) {
-    this.show();
-  }
 
   this.adjust();
 }
 
 FilterPanel.prototype.adjust = function() {
-  if ($('body').width() > 768) {
+  if ($(document).width() > helpers.BREAKPOINTS.LARGE) {
```
ccostino: Given that this line is repeated in the `setHeight` method and may change again in the future, would it make sense to move this conditional into its own method, or do you think that's getting too granular?

js/filter-panel.js
```diff
@@ -25,21 +28,25 @@ function FilterPanel(options) {
...
 
   this.filterSet = new FilterSet(this.$form).activate();
-  if (!_.isEmpty(this.filterSet.serialize())) {
-    this.show();
-  }
 
   this.adjust();
 }
 
 FilterPanel.prototype.adjust = function() {
```
ccostino: Is `adjust` an appropriate name for this method when the end result is showing or hiding the filter panel?  Should this be modifying the width of the panel instead since this appears to be checking for how much screen real estate is available to display the panel?

This is just my observation in processing this mentally, I haven't had a chance to run it yet for myself.

js/filter-tags.js
```diff
@@ -0,0 +1,79 @@
...
+var _ = require('underscore');
+
+var BODY_TEMPLATE = _.template(
+  '<div class="data-container__tags">' +
+    '<h3 class="tags__title">Viewing: ' +
+      '<span class="js-tag-title tags__title__text">{{ title }}</span>' +
+    '</h3>' +
+    '<ul class="tags"></ul>' +
+  '</div>',
+  {interpolate: /\{\{(.+?)\}\}/g}
```
ccostino: I definitely appreciate the handlebars-style delimiters!

js/filters.js
```diff
@@ -115,9 +145,27 @@ function TypeaheadFilter(elm) {
   var key = this.$body.data('dataset');
   var dataset = typeahead.datasets[key];
   this.typeaheadFilter = new typeaheadFilter.TypeaheadFilter(this.$body, dataset);
```
ccostino: Just to be clear on this line and the intent here (because the `new` keyword in JS always jumps out at me), this is deliberately setting `this.typeaheadFilter` to be the instance of the object created and returned by `typeaheadFilter.TypeaheadFilter` itself, right?

This makes sense in the context of the next line, but I just wanted to be sure that was the intent here.

ccostino:
@jmcarp:  Only a handful of things caught my eye initially in just reading through the code; it's well thought-out and organized and I think it addresses the two-way binding between the filters and filter tags well.  And thank you for using `self` instead of `that` (much more descriptive!) when storing a reference to `this`. :-)

I still need to get the repos setup and running locally and then I'll run the code and tests myself, but I wanted to get some initial feedback in place while these questions were fresh in my mind.

js/filters.js
```diff
@@ -115,9 +145,27 @@ function TypeaheadFilter(elm) {
...
+  this.typeaheadFilter.$body.on('change', 'input[type="checkbox"]', this.handleNestedChange.bind(this));
 }
 
 TypeaheadFilter.prototype = Object.create(Filter.prototype);
 TypeaheadFilter.constructor = TypeaheadFilter;
 
+TypeaheadFilter.prototype.handleChange = function() {};
```
ccostino: What is the purpose of having this in here if it's an empty method?

js/filters.js
```diff
@@ -115,9 +145,27 @@ function TypeaheadFilter(elm) {
...
+
+TypeaheadFilter.prototype.handleNestedChange = function(e) {
```
ccostino: Just curious, why is this in here as opposed to typeahead-filter.js (or vice versa, why are the other methods defined in typeahead-filter.js)?

js/filters.js
ccostino: Come to think of it, why is this setup this way as opposed to the way the other objects are setup?  Is it because the Typeahead filter itself is being defined in here with the other filter types, but its behavior is being defined in `typeahead-filter.js`?

ccostino:
I've added a couple of other remarks (mostly questions around code structure at this point), and for what it's worth when I run `npm test` two errors appear:

```
Chrome 48.0.2564 (Mac OS X 10.10.5) filter panel should start off closed FAILED
	AssertionError: expected true to be false
	    at expectClosed (/var/folders/15/brz9h_6d2zj753rcprpw25yr0000gp/T/5104d76665af08f3f18159455d04e309.browserify:44733:29 <- tests/filter-panel.js:23:0)
	    at Context.<anonymous> (/var/folders/15/brz9h_6d2zj753rcprpw25yr0000gp/T/5104d76665af08f3f18159455d04e309.browserify:44760:5 <- tests/filter-panel.js:50:0)
Chrome 48.0.2564 (Mac OS X 10.10.5) filter panel should toggle on click FAILED
	AssertionError: expected false to be true
	    at expectOpen (/var/folders/15/brz9h_6d2zj753rcprpw25yr0000gp/T/5104d76665af08f3f18159455d04e309.browserify:44725:29 <- tests/filter-panel.js:15:0)
	    at Context.<anonymous> (/var/folders/15/brz9h_6d2zj753rcprpw25yr0000gp/T/5104d76665af08f3f18159455d04e309.browserify:44765:5 <- tests/filter-panel.js:55:0)
```

But I see the tests all pass here in the PR, and I haven't been able to track down any actual issue yet.

js/filter-tags.js
```diff
@@ -0,0 +1,79 @@
...
+    .on('filter:renamed', this.renameTag.bind(this));
+
+  this.$list.on('click', '.js-close', this.removeTagDom.bind(this));
+}
+
+TagList.prototype.addTag = function(e, opts) {
+  this.removeTag(opts.key, false);
+  this.$title.html('');
+
+  var $tag = TAG_TEMPLATE(opts);
```
ccostino: Small nitpick:  looks like this could be moved to the beginning of this method (accounting for JS variable hoisting and code readability/consistency).

js/filter-panel.js
jmcarp: We can probably come up with a clearer name for this method--maybe something like `setInitialDisplay`. I think the behavior is what @noahmanger intended--we want the panel to default to open on big screens and closed on small screens, with no intermediate widths.

js/filters.js
jmcarp: This is basically a historical accident. The `TypeaheadFilter` class was written well before we added the `Filter` and `FilterSet` abstractions--so rather than rewrite the earlier code, this class just wraps a nested legacy typeahead. It doesn't look like the nested object is ever used on its own (although I guess it could be at some point), so I think it would make sense to merge the two classes. But I would defer this to a separate patch, since this one's focused on adding filter tags.

js/filters.js
jmcarp: This would go away if we decide to merge the two typeahead classes later on. But for now, I wanted to have `filter:added` and `filter:removed` events only emitted by `Filter` and its subclasses.

js/filters.js
jmcarp: This overrides the inherited `Filter.prototype.handleChange` method, which doesn't work well for this filter subclass--the base class calls `handleChange` when any `input` changes, but in this case, we want to ignore changes on the text box and only listen on the checkboxes. Definitely open to suggestions about a better way to handle this, though.

js/filters.js
ccostino: Ah okay, that makes sense, thanks for the clarification!  +1 on the separate patch for any work involved in merging the two.

js/filters.js
ccostino: G'ah, didn't make the connection from the Filter object, sorry.  Off the top of my head I can't think of a better way to handle this (I would've done the same), but it's something I'll look into more to see if there are more appropriate solutions out there.

PR #216, opened 2015-12-24
Adds new terms

emileighoutlaw:
Happy holidays: the gift of more glossary terms!

cc @noahmanger ?

### DIFF
--- a/js/terms.json
+++ b/js/terms.json
@@ -133,7 +133,7 @@
   },
    {
     "glossary-term": "Separate segregated fund (SSF)",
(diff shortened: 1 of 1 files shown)


noahmanger:
:fireworks: Yes! This is great.

Fixed a couple small things to keep an eye out for in the future:
- Make sure there's a comma between every term/definition pair (right after the closing `}`) _but_ not after the last one of the entire list
- Make sure to escape quotation marks used within definitions: `\"this term is in quotes\"`

emileighoutlaw:
Ah, yes! Will do. Thanks @noahmanger!

PR #217, opened 2015-12-30
Tooltip styling

jenniferthibault:
### Changes:

**Previous:**
Font sizes are inconsistent (some places 1.6, some 1.4rem, padding is too tight and has different values in different places.
screen shot 2015-12-30 at 10 56 55 am
screen shot 2015-12-30 at 11 06 48 am

**Updated:**
Font sizes are made consistent @ 1.4rem, padding is set universally to 1.5rem, increased line height for legibility.
screen shot 2015-12-30 at 11 05 47 am
screen shot 2015-12-30 at 11 07 15 am

### Review:
@noahmanger

### DIFF
--- a/scss/components/_tooltips.scss
+++ b/scss/components/_tooltips.scss
@@ -17,7 +17,7 @@
   border: 2px solid $primary;
   background-color: $cream;
   font-family: $sans-serif;
-  padding: .3rem .5rem;
+  padding: 1.5rem;
   position: absolute;
   text-align: center;
   z-index: $z-tooltip;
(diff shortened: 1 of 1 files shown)


noahmanger:
Just checking, is this what you want? I think there might be double padding here:

screen shot 2016-01-04 at 5 35 50 pm

jenniferthibault:
Wah! No! That does look like double. Which element should actually have the padding applied? The tooltip content, or the tooltip container?

noahmanger:
:+1: I figured. I removed it on the tooltip content.

jenniferthibault:
thank you!

PR #218, opened 2016-01-04
Styling calendar

jmcarp:
* Miscellaneous cleanup
* Make typeahead filters (kind of) work

Note: This implementation of typeahead filter tags shows candidate or committee IDs, rather than names. That's because when we load a page with IDs in the URL, we don't have the names right away--we have to get those via AJAX. While we're waiting, we show "Loading..." instead of the name by the checkbox elements. We *could* update both tags and checkbox labels on AJAX load, but that seems like overkill--it might be simpler just to hide the checkboxes entirely, then render on load.

Also, while I was looking at this patch, I noticed that some fields create ambiguous tags--date and money range filters don't distinguish between minimum and maximum values, for example.

### DIFF
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -70,10 +70,8 @@ FilterSet.prototype.handleTagRemove = function(e) {
 
   if (type === 'checkbox' || type === 'radio') {
     $input.attr('checked', false).trigger('change');
-  }
-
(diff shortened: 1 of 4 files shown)


noahmanger:
Great. Thanks for the cleanup.

PR #221, opened 2016-01-06
Adjusting download spacing

noahmanger:
- This ensures that download buttons line up nice and neatly on desktop
- Fixes a z-index issue where the downloads panel overlapped the filter
button incorrectly on mobile
- Tweaks some styles on small screens

screen shot 2016-01-06 at 10 09 27 am


jmcarp:
Looking good!

PR #226, opened 2016-01-21
Replace event emitter with custom DOM events.

jmcarp:
Two motivations for this change:

* Prefer native browser functionality over third-party modules
* Using eventemitter requires that all modules that share events use the same event dispatcher, which is all right for now but could get complicated if we continue splitting off components into separate repos.

Ping @noahmanger

### DIFF
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -65,7 +65,7 @@ FilterSet.prototype.clear = function() {
 
 FilterSet.prototype.handleTagRemove = function(e, opts) {
   var $input = this.$body.find('#' + opts.key);
-  var type = $input.attr('type') || 'text';
+  var type = $input.get(0).type;
 
   if (type === 'checkbox' || type === 'radio') {
     $input.attr('checked', false).trigger('change');
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -5,7 +5,6 @@ var _ = require('underscore');
 var URI = require('urijs');
 
 var Filter = require('./filters').Filter;
-var events = require('./events');
 
 var KEYCODE_ENTER = 13;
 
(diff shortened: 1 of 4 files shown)


noahmanger:
Looking good, we should just remove it from package.json too.

jmcarp:
Not yet--we're still using it for the accordion and typeahead modules.

js/filter-set.js
```diff
@@ -64,9 +63,9 @@ FilterSet.prototype.clear = function() {
   });
 };
 
-FilterSet.prototype.handleTagRemove = function(e) {
-  var $input = this.$body.find('#' + e.key);
-  var type = $input.attr('type');
+FilterSet.prototype.handleTagRemove = function(e, opts) {
+  var $input = this.$body.find('#' + opts.key);
+  var type = $input.attr('type') || 'text';
```
noahmanger: Why set a default value here?

jmcarp: Because "text" is the default type on an `<input>`. If we don't specify a default and ever use a plain `<input>` without a `type` attribute, this code will fail. We could drop the default if we want to require that all `<input>` elements in a `FilterSet` specify a `type`.

noahmanger: Ah gotcha. I didn't realize that you could actually leave the `type` off an `input`.

jmcarp: Actually, I'm proposing a little change that handles this a bit better. It turns out that `$input.attr('type')` gives the value of the `type` as encoded in the source, which means it's `undefined` if we didn't provide a `type` for some element. But using the native DOM interface works here, since it gives us the type that the element has taken on, either because it was set explicitly or because it was the default value: `$input.get(0).type`.

PR #231, opened 2016-01-27
Consolidate typeahead

jmcarp:
Branched off `styling-calendar`.

To be determined:
* [x] Should all typeahead filters support search by ID and free text?
* [x] How to clarify the behavior of each filter (ID vs free text)

cc @noahmanger @LindsayYoung @jenniferthibault

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -143,8 +143,9 @@ function TypeaheadFilter(elm) {
   Filter.call(this, elm);
 
   var key = this.$body.data('dataset');
+  var allowText = this.$body.data('allow-text') !== undefined;
   var dataset = typeahead.datasets[key];
-  this.typeaheadFilter = new typeaheadFilter.TypeaheadFilter(this.$body, dataset);
+  this.typeaheadFilter = new typeaheadFilter.TypeaheadFilter(this.$body, dataset, allowText);
   this.typeaheadFilter.$body.on('change', 'input[type="checkbox"]', this.handleNestedChange.bind(this));
 }
 
--- a/js/filters.js
+++ b/js/filters.js
@@ -151,6 +151,13 @@ function TypeaheadFilter(elm) {
 TypeaheadFilter.prototype = Object.create(Filter.prototype);
 TypeaheadFilter.constructor = TypeaheadFilter;
 
+TypeaheadFilter.prototype.fromQuery = function(query) {
+  var values = query[this.name] ? ensureArray(query[this.name]) : [];
+  this.typeaheadFilter.getFilters(values);
+  this.typeaheadFilter.$body.find('input[type="checkbox"]').val(values);
+  return this;
+};
+
 TypeaheadFilter.prototype.handleChange = function() {};
 
 TypeaheadFilter.prototype.handleNestedChange = function(e) {
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -22,17 +22,22 @@ function FilterSet(elm) {
 
 FilterSet.prototype.activate = function() {
   var query = URI.parseQuery(window.location.search);
-  this.filters = _.chain(this.$body.find('.js-filter'))
-    .map(function(elm) {
-      var filter = Filter.build($(elm)).fromQuery(query);
-      return [filter.name, filter];
-    })
-    .object()
-    .value();
-  this.fields = _.chain(this.filters)
-    .pluck('fields')
-    .flatten()
-    .value();
+  if (_.isEmpty(this.filters)) {
+    this.filters = _.chain(this.$body.find('.js-filter'))
+      .map(function(elm) {
+        var filter = Filter.build($(elm)); // .fromQuery(query);
+        return [filter.name, filter];
+      })
+      .object()
+      .value();
+    this.fields = _.chain(this.filters)
+      .pluck('fields')
+      .flatten()
+      .value();
+  }
+  _.each(this.filters, function(filter) {
+    filter.fromQuery(query);
+  });
   return this;
 };
 
--- a/js/typeahead-filter.js
+++ b/js/typeahead-filter.js
@@ -6,19 +6,25 @@ var $ = require('jquery');
 var URI = require('urijs');
 var _ = require('underscore');
 
+var ID_PATTERN = /^\w{9}$/;
+
+function slugify(value) {
+  return value
+    .trim()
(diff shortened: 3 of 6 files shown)


jmcarp:
Assigning this to @noahmanger to coordinate decision-making around filter behavior, update the issue once choices get made, and possibly review changes once everything's implemented.

noahmanger:
Thanks for getting this started. I think there's a pretty strong argument that we'll want certain filters to *only* accept typeahead, so I wanted to try to figure out how the two types would be distinguished. 

typeahead

This is a pretty simple solution. I imagine there's maybe more we could do to differentiate, but thought we could start here.

Of course, if we figure that some of the fields could actually accept free text, then this isn't a problem, but worth exploring either way.

@jenniferthibault what do you think?

jenniferthibault:
Before I amend this with a small behavioral addition, I want to make sure I'm not off-base. @noahmanger is it possible that for a type 1 search, a user could start typing "bill", see all of the below typeahead options, and hit enter to search for "bill" as plaintext?

Or is the behavior that you can only enter a basic text string IF (and only if) there is no matching or suggested typeahead options?

noahmanger:
The first one: you can hit enter and search just text (or click the bottom part of the dropdown).

jenniferthibault:
The functionality is totally in line with how I expected the behavior to work, but I'd like to suggest a few text changes to make language more concise.

`Matching committees` >> `Committees`
`Search for all contributors named: “bill” ` >> `"bill"` ; and move this to the first option under the search box, instead of last. This is a more common pattern, and I think logical to keep them visually grouped.
Defer entirely to @emileighoutlaw 's judgement on microcopy here.

And some styling tweaks:
- the darker shade should act as a highlight, and default to the first suggested option, whether that's `"word"` or option in a typeahead, so that when a user hits enter, it selects what is highlighted as the query
- is it possible to highlight (or in this design, make bold) the part of the word matching the text entry in the results? If bold isn't distinct enough, we could consider a text highlight.
- try not to use italics
- small indent on the list of committees under the matching results header
- shifting placement of the search icon to the end of the bar so that search text can remain aligned with results

Here's what that looks like, with Noah's original above for easy comparison: (better viewing experience if you open in a new tab)
typeahead-type 1
typeahead-type 2

emileighoutlaw:
I love Jen's microcopy suggestions. 

I have one question about the Committee ID field. I love the note added for clarity, but even with it, `Committee ID` only looks like a place to enter a committee ID to me. What do you think about something more like `Search for a committee's ID` or `Find a committee's ID` ?

noahmanger:
@emileighoutlaw great question, and that's exactly what we're struggling with. That sounds like it could work.

@jenniferthibault love the modifications. Only question I have is if the search icon button on the right is meant as a button and if so, what does it do?

jenniferthibault:
Nope, not meant as a button, just as an icon that brings more attention to the field. It wouldn't have any interaction. Similar to Google's search field here, all the interaction happens through typeahead or when the user hits enter to commit the term to search.

https://www.google.com/#q=law+and+order
screen shot 2016-01-31 at 8 05 09 pm

noahmanger:
Cool cool. So, even in the google case it's still a button, but I guess it doesn't really serve a function other than just being there as a safety net for folks. 

@jmcarp I'd say were good to go on continuing implementation based on Jen's mocks above. The latest work on the new datatable columns and filters suggests that having a way to require a specific candidate / committee ID will be helpful, so I'd like to at least get this to a point where can do some quick and dirty usability testing and see if people get too hung up before we try anything else.

jenniferthibault:
On the search bar, right! A dummy/safety button of sorts. Same as we're using here:
screen shot 2016-02-01 at 10 11 42 pm

I was thinking of this existing on our site, but should have added this screenshot the first time around.

jmcarp:
I pushed some updates to fec-style/consolidate-typeahead and openfec-web-app/consolidate-typeahead that behave like @jenniferthibault's designs (I haven't addressed anything text-related). To test it out, verify that the contributor filter allows text search on the receipts data table page, but the committee filter doesn't. Reminder: use https://github.com/18F/openFEC/pull/1462 to get back the expected results from the API.

@noahmanger: when you have time, could you either check out the relevant branches and review or do a quick screenshare demo / critique?

noahmanger:
Ok, I've got this styled enough to test:

image

I couldn't get it to be the full-width of the input + button without some annoying work, so let's see how this fares in testing and then we can refine later.

jenniferthibault:
1000% ok with this for testing. And actually, I'd prefer to hold off on refining designs in #240 until this is tested and we have some interaction feedback. Is that ok?

emileighoutlaw:
Is it possible to change that microcopy to read `Enter a name or committee ID` ? I think that's just a little more concise!

jmcarp:
This is part of the consolidated typeahead feature. Re-assigning to @ccostino for review.

js/filters.js
```diff
@@ -143,14 +143,22 @@ function TypeaheadFilter(elm) {
   Filter.call(this, elm);
 
   var key = this.$body.data('dataset');
+  var allowText = this.$body.data('allow-text') !== undefined;
   var dataset = typeahead.datasets[key];
-  this.typeaheadFilter = new typeaheadFilter.TypeaheadFilter(this.$body, dataset);
+  this.typeaheadFilter = new typeaheadFilter.TypeaheadFilter(this.$body, dataset, allowText);
```
ccostino: One quick note about this - this value is going to be the result of the evaluation of `this.$body.data('allow-text') !== undefined`, not the actual value of `this.$body.data('allow-text'` itself.  I just wanted to be sure that is the intent here and not to use the actual value of the `data-allow-text` attribute.

ccostino: Or, should this be expressed differently, e.g.:  `this.$body.data('allow-text') !== undefined ? this.$body.data('allow-text') : false;`?

jmcarp: The intention here is that `data-allow-text` is either present or absent:

```
<div data-allow-text>text allowed</div>
<div>text not allowed</div>
```

The `allowText` variable in the javascript is meant to be a boolean flag that signals whether `data-allow-text` is set or not.

ccostino: Okay, so the presence of `data-allow-text` signifies `true` regardless of what it's set to and its absence is the `false` condition for `allowText`.  Just wanted to be sure!

PR #238, opened 2016-02-03
Styling calendar

noahmanger:
Adds simple toggle style for the category / chronological toggle on the calendar:
screen shot 2016-02-02 at 9 08 15 pm

Also:
- Sets a min-height on calendar rows so they're all the same height.
- Moves the `eq` handlebars helper to fec-style

Goes with https://github.com/jmcarp/fec-cms/pull/4

### DIFF
--- a/js/dropdowns.js
+++ b/js/dropdowns.js
@@ -1,9 +1,9 @@
 'use strict';
 
-/* global require, module, document */
-
 var $ = require('jquery');
-var perfectScrollbar = require('perfect-scrollbar/jquery')($);
+require('perfect-scrollbar/jquery')($);
+
+var listeners = require('./listeners');
 
 var KEYCODE_ESC = 27;
 var KEYCODE_ENTER = 13;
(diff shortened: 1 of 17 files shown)


js/listeners.js
```diff
@@ -0,0 +1,26 @@
...
+
+Listeners.prototype.on = function(elm) {
+  var $elm = $(elm);
+  var args = _.toArray(arguments).slice(1);
+  this.listeners = this._listeners || [];
+  this.listeners.push({$elm: $elm, args: args});
+  $elm.on.apply($elm, args);
+};
+
+Listeners.prototype.off = function() {
```
ccostino: Is the intention here to always just shut off all listeners and not provide a way to shut off an individual listener, like how you can add a new one?

js/dropdowns.js
```diff
@@ -1,9 +1,9 @@
 'use strict';
 
-/* global require, module, document */
-
 var $ = require('jquery');
-var perfectScrollbar = require('perfect-scrollbar/jquery')($);
+require('perfect-scrollbar/jquery')($);
```
ccostino: Just to make sure I'm understanding this correctly, this is providing the perfect-scrollbar library to the jQuery object itself ($) by using the `require` statement in this fashion?

jmcarp: It's the other way around--we initialize the perfect-scrollbar plugin by passing it the global jQuery object. Incidentally, use of this library predates me on the project, so @noahmanger can field questions about what it actually does. The only thing I'm doing here is taking out an unnecessary variable declaration that was throwing a jshint error.

js/listeners.js
jmcarp: The purpose of this class is to keep track of event listeners so that they can be deleted in a batch, typically when the `destroy` method of a parent class gets called. We could allow this method to take optional arguments instead, but we haven't run into a use case for that yet. Or we could rename the method to something like `clear`.

ccostino: Okay yes, I saw that above here:  https://github.com/18F/fec-style/pull/238/files#diff-55ad661871cd27444f58e59f424ed11bR133 - this makes sense since with the object destroyed, there's no point in having any of the listeners remain for it.

The naming is what threw me off (pun not intended) initially; `clear` sounds a lot better to me in this case because that makes it... clear as to what's happening. :-)

js/dropdowns.js
ccostino: D'oh!  Not sure why I got that flipped around, but thank you for clarifying.  Basically I was just double checking to see that it truly didn't need to be referenced directly in the code, but it sounds it doesn't.

ccostino:
Aside from the `off` to `clear` renaming in listeners.js (which is done) this looks good to go to me!  Is there anything else that needs to change with regards to a version number bump, like updating the `package.json` file with the version number `1.4.2`?  Or does this happen separately?

PR #247, opened 2016-02-12
Fixing download spacing and truncating download title

noahmanger:
Ping @jmcarp

### DIFF
--- a/scss/components/_downloads.scss
+++ b/scss/components/_downloads.scss
@@ -40,8 +40,10 @@
 }
 
 .download__name {
+  @include u-truncate();
   display: block;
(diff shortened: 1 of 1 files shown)


jmcarp:
Maybe this is an improvement, but truncating the download names makes it a little harder to tell which download is which in the progress panel. For example, all downloads of electioneering data are going to appear to have exactly the same name. Of course, our current system of naming downloads by start dates isn't particularly informative.

screen shot 2016-02-12 at 12 42 46 pm

noahmanger:
Mmm I think it's ok for now.

PR #248, opened 2016-02-12
Add unstyled election filter.

jmcarp:
Needs revisions and tests--propose to defer until after usability testing.

cc @noahmanger


noahmanger:
+1 to tests coming after usability testing. As can style refinements.

@ccostino do you mind taking a look at this? I think best to have you as the primary reviewer on JS from here on out.

ccostino:
Sure thing!  That sounds good to me too with the JS reviews, I'm happy to do that.

ccostino:
Let me know when I should take a closer look.  As a start, this looks to be sound so far.

js/filters.js
```diff
@@ -177,4 +183,73 @@ TypeaheadFilter.prototype.handleNestedChange = function(e) {
...
+  }.bind(this));
+  bins.unshift({
+    name: this.name,
+    cycle: election,
+    min: election - this.duration + 1,
+    max: election,
+    full: true
+  });
+  this.$cycles.html(cyclesTemplate(bins));
+  this.$cycles.find('input').eq(0).prop('selected', true).change();
```
noahmanger: Small thing, but in other toggles we use the attribute "checked". Not sure if you care. Looks like it does the same thing.

js/templates/election-cycles.hbs
```diff
@@ -0,0 +1,15 @@
+{{#each this}}
+  <label>
```
noahmanger: Label should have a `for=[id of input]` attribute and then the `input` needs the same id.

noahmanger:
Apart from my small notes, two behavior questions:

1. When the select is first changed, what is the default time period? Is it the full cycle? If so, it doesn't look like that gets its `selected`/`checked` property set.
2. Do you want to add the filter tags functionality in a subsequent patch or include it in here? I'm ok with either.

jmcarp:
@noahmanger I updated the markup and fixed selections on election year change. Seems reasonable to implement filter tags and styling in separate patches.

noahmanger:
:+1: I'll make a new issue for adding the tags.

PR #250, opened 2016-02-17
Fix failing glossary tests.

jmcarp:
Looks like the build on master broke a few commits ago. This patch fixes the failing tests. Note: please be careful not to merge pull requests with failing tests into master!


noahmanger:
Gah. Sorry about that. Thanks.

PR #254, opened 2016-02-24
Making the bright red color slightly darker for 508 reqs

noahmanger:
CC @jenniferthibault

### DIFF
--- a/scss/_variables.scss
+++ b/scss/_variables.scss
@@ -14,7 +14,7 @@ $ivory: #f5f0e4;
 $cream: #ffffff;
 
 // Accent colors
-$bright-red: #dd3a00;
+$bright-red: #dc3b00;
 $deep-orange: #e36641;
 $deep-red: #961a16;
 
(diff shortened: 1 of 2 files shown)


jenniferthibault:
@noahmanger we need to also change the card text (which I think might be `.slab--secondary` ? ) from `$ivory` (#f5f0e4) to the `$cream` value, which is already set to #fff . Changing the color value of the red alone doesn't meet 4.5:1 contrast

noahmanger:
Oh! Good catch. Thanks.

On Wed, Feb 24, 2016 at 2:30 PM, Jennifer Thibault <notifications@github.com
> wrote:

> @noahmanger <https://github.com/noahmanger> we need to also change the
> card text (which I think might be .slab--secondary ? ) from $ivory
> (#f5f0e4) to the $cream value, which is already set to #fff . Changing
> the color value of the red alone doesn't meet 4.5:1 contrast
>
> —
> <https://github.com/18F/fec-style/pull/254#issuecomment-188485570>.
>



-- 
Noah Manger

jenniferthibault:
Thank you!

PR #256, opened 2016-02-28
Externalize feedback

jmcarp:
Follow-up to https://github.com/18F/openFEC-web-app/pull/1110: delete soon-to be unused feedback and modal code; roll back unused requirements.

Note: the modal component doesn't appear to be used currently, and I'm not sure it was ever used, so I'm deleting it here as well. I also don't remember when we added that code--can you refer me to the PR @noahmanger?


noahmanger:
Yeah, you're correct that the modal code isn't used anywhere. I started it for this, but forgot to delete it when we didn't use it.

noahmanger:
Assigning to @ccostino to review along with the companion web app PR.

PR #257, opened 2016-03-03
fixes capitalization

emileighoutlaw:
Content only. Fixes capitalization in `<li>`.

Thinking this would be an excellent 1-second @jenniferthibault review ;)


jenniferthibault:
:boom: will merge when the tests finish.

PR #261, opened 2016-03-14
Adjusting styles on filters and more

noahmanger:
- Adjusts padding in filter accordions
- Adds style for blue dropdowns
- Fixes margin on mini search bar
- Ensures content on mobile doesn't run up to the edge of the screen


noahmanger:
Conflicts resolved. Whenever we merge this in we should be good to bump the version.

PR #273, opened 2016-03-17
Improve typeahead interaction

noahmanger:
Since the release got pushed back a day, I wanted to take the time to sand off a couple rough edges that were nagging me with the typeahead filters. 

Previously, the only way to select a typeahead filter was by clicking (or arrowing-down and hitting enter) on an option in the list. With this patch, you can now:

### 1. Hit enter in the input
If the input accepts free text, it turns the text entry into the filter:
image

If it accepts only typeahead suggestions, it uses that:
image

### 2. Click the button 
For free text:
image

For suggestions (you must hit tab in order to complete the autocomplete suggestion):
image

There might still be ways we can improve this, but the more I used it the more this stuck out as things that would be good to fix. It shouldn't hold up launch, but if it's straightforward enough, I think it would be worth including.

@jmcarp or @ccostino can one of you take a look?

### DIFF
--- a/js/typeahead-filter.js
+++ b/js/typeahead-filter.js
@@ -36,12 +36,17 @@ var textDataset = {
 var TypeaheadFilter = function(selector, dataset, allowText) {
   this.$body = $(selector);
   this.dataset = dataset;
+  this.allowText = allowText;
 
   this.$field = this.$body.find('input[type="text"]');
   this.fieldName = this.$body.data('name') || this.$field.attr('name');
+  this.$button = this.$body.find('button');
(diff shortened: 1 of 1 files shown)


ccostino:
@noahmanger, I can take a look this morning and get this reviewed.  I had thought about these things as well when I was reviewing other bits of functionality earlier.  Good to see them here! :-)

jenniferthibault:
:clap:

ccostino:
@noahmanger, this looks good!  I'll update the version and publish it after merging, then submit PRs for the CMS and web-app to pin to just the minor version instead of the patch level as we discussed.  Thanks!

Issue #264, opened 2016-03-17
To make the site feel less stale and increase color legibility, recommend replacement colors for beige


jenniferthibault:
Based on the visual feedback, we learned that over time we were using more of the beige color palette than intended in the original design direction, because the need for neutral explainers or low-priority info panels became really common. Additionally, we've heard that $ivory sometimes appears very different shades on different user computers—sometimes pinkish, sometimes very yellow. Having so much beige around is making the site feel a little stale, and I'd like to look into swapping it out for another neutral palette complement. 

I'll be looking specifically at instances of these colors:

<img width="637" alt="screen shot 2016-03-17 at 1 47 55 pm" src="https://cloud.githubusercontent.com/assets/11636908/13855458/f9ec2e80-ec46-11e5-8276-042c3b4d5878.png">

<img width="647" alt="screen shot 2016-03-17 at 1 47 46 pm" src="https://cloud.githubusercontent.com/assets/11636908/13855460/fe31bcda-ec46-11e5-9a78-7a6dc0273050.png">

noahmanger:
Resolved by https://github.com/18F/fec-style/pull/279

PR #279, opened 2016-03-23
Color testing

jenniferthibault:
Still very much working through this! Try not to fall out of your chair laughing at my commit attempts. I'm hoping @noahmanger can help clean up any damage or unnecessary code I've left in the wake once this is done.

This primarily addresses concerns seen in #264. I've integrated the neutral grey palette from the U.S. Web Design standards, which I believe maintains the neutrality we need for these elements, but blends in a bit better.

screen shot 2016-03-23 at 4 39 20 pm

screen shot 2016-03-23 at 4 41 00 pm

A small but noticeable change for the tone across all parts of the site, is that I've changed the default text color from federal blue to a dark charcoal (nearly black) color. I think this will help a lot on the calendar and registration and reporting side, and will help pop the content from the structure more.

Areas I still need to look at to make sure this will be a viable set of changes:
- everything on the CMS-side (need to spin up a local instance)
- glossary (going to pair up with Noah to make sure I wrangle the styles correctly)
- notifications & alerts


scss/_variables.scss
```diff
@@ -7,11 +7,11 @@ $base-font-family: $serif;
 $crimson: #631010;
 $orange: #f77b42;
 
-$federal-blue: #022c53;
+$federal-blue: #112e51; /* U.S. Web Design Standards color primary-darkest */
```
noahmanger: This is a good catch. @jenniferthibault can you change comments throughout to be: `// Comment` rather than: `/* Comment */` ? These comments will be ignored in the compiled CSS, which keeps the file size smaller overall.

jenniferthibault:
Ok! I think this is close, and ready for your eyes & assistance @noahmanger.

There are a few bits I made weird along the way. I'm not saying "broken" :)

This is part of a three-piece PR set:
https://github.com/18F/fec-cms/pull/257
https://github.com/18F/openFEC-web-app/pull/1141

**Glossary:**
 I changed the palette to a universal base dark color, so it matches both the crimson and federal blue palette no matter which page it's opened on. Somehow, I accidentally budged the closing X button  out of the corner. I did intentionally move the glossary headline down, but the button should have stayed where it was.

screen shot 2016-03-28 at 6 57 45 pm
screen shot 2016-03-28 at 6 57 57 pm

**Neutral alert message**
Making the filter panel neutral now mad it so that the message background is the same color. I'd like to use the same transparent white background that you see on error messages for the feedback button, but I couldn't quite get it working.

screen shot 2016-03-28 at 5 59 48 pm
screen shot 2016-03-28 at 6 00 55 pm
 
**Dark slab**
I was trying to make a dark neutral-colored slab to help reduce some of the orange friction. I couldn't figure it out enough to get into a great place, but here's what I gout do through basic inspector color changes.

screen shot 2016-03-28 at 6 43 15 pm

screen shot 2016-03-28 at 7 10 38 pm

noahmanger:
Looks like you're trying to use a `$gray-darkest` variable, but it's undefined. Do you want to use that, or `$gray-dark`?

jenniferthibault:
Ahh that's definitely at least part of it. I meant it to be `$gray-dark`, I'll send that patch now.
I wasn't sure what `u-form-color` did here, so I think I mostly left it, if it's an important to switch, what does it impact?

noahmanger:
Well, the Sass compilation crashes when there's undefined variables. So I
can't even build the styles without that.

On Mon, Mar 28, 2016 at 4:39 PM, Jennifer Thibault <notifications@github.com
> wrote:

> Ahh that's definitely at least part of it. I meant it to be $gray-dark,
> I'll send that patch now.
> I wasn't sure what u-form-color did here, so I think I mostly left it, if
> it's an important to switch, what does it impact?
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/18F/fec-style/pull/279#issuecomment-202626587>
>



-- 
Noah Manger

noahmanger:
Ok, fixed the first couple things. Lets do the dark slab in a separate change. I have other thoughts too, but those too can come next.

PR #285, opened 2016-03-28
Adding required signifier to feedback form

noahmanger:
Resolves https://github.com/18F/FEC/issues/270

Note, I moved the "*Required" note to below the disclaimer paragraph, rather than above:

image


emileighoutlaw:
Looks good! I like the smaller font on this, too

PR #289, opened 2016-04-04
Adding active style for calendar toggle

jenniferthibault:
Button copy and icons should be `$inverse` when active.

I couldn’t get this to show up locally, so no screenshot, and it might mean I didn’t entirely do it correctly :( 

Review @noahmanger


noahmanger:
This is addressed in the button revamp, so closing this.

PR #298, opened 2016-04-07
fec-style specific contributing guidelines

adborden:
As discussed in our fec-style meeting on 2015-03-25.
https://docs.google.com/document/d/1H-_P6nX4WRoZfCeZ_jJcF5lg0k5BgPfwvnUhrBsBDwo/edit

I started by adding two lines to `CONTRIBUTING.md`, but I'm afraid it will get lost since there is a ton of content in there covering all of our projects. I know @ccostino  and @emileighoutlaw are re-doing some of the `README` content, but not sure if the `CONTRIBUTING` docs are part of that.

I've never used the github templates before, but figured this would be a good chance to try them out so that our `fec-style` contributing guidelines are in your face when submitting a PR.

cc @noahmanger @jenniferthibault

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,11 @@
+## Summary
+
+_Include a summary of all work and changes._
+
+## Screenshots
+
+- _Update the styleguide documentation, include a screenshot if applicable._
+- _Include a screenshot of the new/updated styles in context (“in the wild”)._
+
+
+_cc @username_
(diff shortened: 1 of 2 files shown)


noahmanger:
I like this! Good idea to use the PR template. Let's give it a go.

ccostino:
Thanks, @adborden!  The work @emileighoutlaw and I are doing would eventually encompass things like this but we're very happy to have the help. :-)  As we continue to have conversations around our work style and process, I think the `CONTRIBUTING` doc makes the most sense to write down whatever we decide to do (and this will go across repos, not just here in `fec-style`).

Issue #297, opened 2016-04-06
Implement new button styles


noahmanger:
Implement the new button styles defined in https://github.com/18F/fec-style/issues/288.

@jenniferthibault , let me know when this is good to move to implementation.

jenniferthibault:
Here's [an InVision](https://invis.io/Z46U2A16W) with my take on how the styles from #288 would be implemented across screens: I've color-coordinated style matches for toggles, buttons, and drop-down lists. Feedback welcome ( @adborden @edmullen ! ) though I haven't done much annotating in the mockups, I could do that tomorrow if needed.

(Sorry, screenshots are a bit scrappy)

Closing out #228 , #236,  and #235 in favor of this inplementation

noahmanger:
This is great! Thanks @jenniferthibault . Really helpful to see how they work in the wild, too. I left a few comments / questions in there that I noticed, but either way I'm going to start on this Thursday.

jenniferthibault:
I wish I could reply to an InVision comment with a photo, but no dice, so here we are. @noahmanger per your first comment about button syntax, it's helpful for me to see the button styles in two ways (thinking to how we decide to present them in the style guide)

1: How many discrete button styles are there overall?
The types of buttons and how or if their style needs to be changed depending on what background they appear on. 
![screen shot 2016-04-07 at 12 15 16 pm](https://cloud.githubusercontent.com/assets/11636908/14358330/6fc427ce-fcba-11e5-933e-31f269355adb.png)


2: How do the buttons get used with different colored backgrounds? (What buttons can I choose from depending on what background color we're using in this instance?)
![screen shot 2016-04-07 at 12 08 18 pm](https://cloud.githubusercontent.com/assets/11636908/14358241/18c7be22-fcba-11e5-9956-0374e8191d77.png)

This makes it clear to me that the standard gray button could be used on the blue and red backgrounds, which you wouldn't know on the first one. (Though given that the solid backgrounds are usually call-to-action type areas, I haven't needed it yet)

noahmanger:
Will we ever use the aqua or orange call to action buttons on neutral backgrounds? Wondering if we can just set those as a rule on certain backgrounds.

jenniferthibault:
Not as I was planning it. The neutral background would be the same combos as a white background, because there is no fill to the button backgrounds (they are transparent as I've build them) Here's what that would look like more explicitly:

[...]

Issue #260, opened 2016-03-11
Refine styling of new election year filter


noahmanger:
Redesign this to look better:

<img width="269" alt="screen shot 2016-03-10 at 4 09 18 pm" src="https://cloud.githubusercontent.com/assets/1696495/13689006/7eb23ab4-e6da-11e5-937e-c0bfeaa4c80e.png">

PR #302, opened 2016-04-08
Button revamp

noahmanger:
## Summary
Implements the new button, input, select and dropdown styles to go with the overall visual refresh. 

## Screenshots
This has effects throughout the web app and CMS, but here's some of the highlights:

In the styleguide:
styleguide example
image
image
image
image

## Issues resolved
- Resolves https://github.com/18F/fec-style/issues/297
- Resolves https://github.com/18F/fec-style/issues/260

## Related PRs
- https://github.com/18F/openFEC-web-app/pull/1164
- https://github.com/18F/fec-cms/pull/267

cc @adborden @jenniferthibault


noahmanger:
Addressed a couple of the Hound warnings and turned off another. The duplicate background-image warning though is ok. We do this to set a png fallback for svgs, but in other circumstances the warning is helpful.

scss/components/_dropdowns.scss
```diff
@@ -52,6 +52,10 @@
     padding: 1rem .5rem 1rem 1.5rem;
     right: .5rem;
   }
+
+  & + .dropdown__panel {
```
adborden: Yeah, this looks like a bug in linter... unless you can start a nested rule with `+ .class`????

jenniferthibault:
A few things in no particularly order:

Some mysterious ghost buttons in the style guide which weren't there last I checked, I'm not entirely sure what's trying to happen there:

screen shot 2016-04-08 at 6 08 34 pm

I think I'm also seeing a missing hover effect on dropdowns/selects. 
gif1

It should have the same hover effect as the button-alt style (in all cases)
gif2

scss/components/_dropdowns.scss
noahmanger: Yeah that doesn't work (tried it).

scss/layout/_grid.scss
```diff
@@ -48,6 +50,21 @@
 
 // BREAKPOINT: LARGE
 @include media($lg) {
+  .grid--3-wide {
+    .grid__item {
+      @include span-columns(4);
+      margin-bottom: 0;
```
adborden: Could the `margin-bottom` be included in a more generic rule? Doesn't seem like it would be specific to `.grid--3-wide` (or `.grid--4-wide`).

noahmanger:
Thanks! I'll add the hover.

The ghost buttons are actually the `.button--alt-primary` and `.button--alt-secondary` classes, which are white outlines with white text, which is why they don't show up here. In the future we can find a better way to show them.

scss/layout/_grid.scss
noahmanger: Not sure. This is currently used in places where we have a 2x2 or 4x1 grid. In the case of the 2x2, you want to have the margin-bottom set at all sizes; and in the case of 4x1, you want it just set at small screens when it becomes a 1x4. 

Not sure what the right way to abstract is. Suggestions?

PR #304, opened 2016-04-09
Styles for legal search results

adborden:
## Summary

Added some minor styles in order to implement the [Legal search results page](https://projects.invisionapp.com/share/5N6NL76QM#/screens/145835130). [Trello](https://trello.com/c/g0MX7L9w/62-hook-ui-up-to-search-api).

- Disabled `side-nav` links
- Datatable suitable for text.
- Classes for adjusting datatable cell widths
- Secondary text for the `results-info` header

## Screenshots

screenshot from 2016-04-08 18-31-18

cc @edmullen @noahmanger @jenniferthibault

### DIFF
--- a/scss/components/_side-nav.scss
+++ b/scss/components/_side-nav.scss
@@ -9,7 +9,7 @@
 //       <li class="side-nav__item"><a class="side-nav__link is-active" href="#">Link (active)</a></li>
 //       <li class="side-nav__item"><a class="side-nav__link" href="#">Link</a></li>
 //       <li class="side-nav__item"><a class="side-nav__link" href="#">Link</a></li>
-//       <li class="side-nav__item"><a class="side-nav__link" href="#">Link</a></li>
+//       <li class="is-disabled side-nav__item"><span class="side-nav__link" href="#">TBD</span></li>
 //     </ul>
 //   </nav>
 // </div>
@@ -29,6 +29,15 @@
 
   &:last-child {
     padding-bottom: 0;
+  }
+
+  &.is-disabled {
+    .side-nav__link {
+      &:hover,
+      &:focus {
+        background-color: transparent;
+      }
+    }
   }
 }
 
--- a/scss/modules/_results-info.scss
+++ b/scss/modules/_results-info.scss
@@ -29,6 +29,11 @@
   margin-top: 1rem;
 }
 
+.results-info__secondary {
+  line-height: 2;
+  font-family: $sans-serif;
+}
+
 .results-info--top {
   border-top: none;
 }
--- a/scss/modules/_results-info.scss
+++ b/scss/modules/_results-info.scss
@@ -29,7 +29,7 @@
   margin-top: 1rem;
 }
 
-.results-info__secondary {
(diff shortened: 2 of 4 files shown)


jenniferthibault:
@noahmanger this has brought up a design pattern question for me. 

One half of my brain is saying "That 'Hit' text is a big block of AO paragraph text, so it should be Gandhi Serif"
but the other half is saying "but it's in a table, (and it's sort of acting like data) so I understand why it's Karla" 

screen shot 2016-04-11 at 6 06 34 pm

Where do your eyes and brain fall on that?

I'm leaning that at the very least, if possible, the Hit copy that isn't a title (currently regular weight) should be Gandhi Serif, but would love to hear your thoughts. 

(other small thing, the dropdown style in the search bar needs to be applied so that the text shows up)

scss/modules/_results-info.scss
```diff
@@ -29,6 +29,11 @@
   margin-top: 1rem;
 }
 
+.results-info__secondary {
```
noahmanger: We've been using `primary` and `secondary` to mean the blue and crimson backgrounds, respectively. For clarity, what about changing this to something like `.results-info__details` or something?

scss/components/_side-nav.scss
```diff
@@ -30,6 +30,15 @@
   &:last-child {
     padding-bottom: 0;
   }
+
+  &.is-disabled {
```
noahmanger: Small thing: You shouldn't need this if you just don't wrap the inner element in a `.side-nav__link` class. You could just leave it `<l class="is-disabled side-nav__item"i>TBD</li>` (which is what we do elsewhere.

I don't mind having this in there if that's the way you'd rather go though.

adborden: I tried removing the inner span because that sounds like the cleaner way to go, but looks like the link style is providing some padding. Is that a bug? What do you suggest?

screenshot from 2016-04-11 16-32-00

noahmanger: Ah! My bad. I was looking at a similar but different pattern:
image

I think what you have is the right way to go.

scss/modules/_results-info.scss
adborden: Fixed.

jenniferthibault:
@edmullen thank you for this! 

What I'm seeing more clearly now is that our search results patterns on the data side aren't mature enough for the context of the legal stuff search results. In a lot of ways, the legal resources search patterns are setting the benchmark for search across the whole betaFEC. 👍 @edmullen 

I'd like to update the styles of the search pattern with the needs of the more complex elements of legal results front of mind, but this is likely a task for the next sprint. I don't think it's a big lift,(probably just typography styles and maybe sizing adjustments) but I would like the chance to see data side search results and legal side search results patterns together before we alter styles.

** Until I can work that into the sprint planning, can we move fwd with the middle option mocked up above @adborden? Gandhi body text?

noahmanger:
LGTM. @adborden can you just change the body text to serif like @jenniferthibault mentioned above then I'll merge?

PR #307, opened 2016-04-13
Breadcrumbs

noahmanger:
## Summary

Adds styles for the breadcrumbs as designed in https://github.com/18F/openFEC/issues/1595

I know that issue isn't 100% done, but it will be easy to make changes to this once it is.

## Screenshots

image

image

cc @adborden @jenniferthibault 

## To do still
- [x] Set max-width and truncate items

### DIFF
--- a/scss/components/_breadcrumbs.scss
+++ b/scss/components/_breadcrumbs.scss
@@ -24,25 +24,34 @@
 // Styleguide components.breadcrumbs
 
 .breadcrumbs {
-  color: $gray;
-  float: left;
-  font-family: $sans-serif;
-  letter-spacing: -.3px;
+  display: none;
 }
 
-.breadcrumbs__item {
-  float: left;
-  padding: .8rem 0;
-}
+@include media($lg) {
+  .breadcrumbs {
+    color: $gray;
+    display: block;
+    float: left;
+    font-family: $sans-serif;
+    letter-spacing: -.3px;
+  }
 
-.breadcrumbs__item--current {
-  color: $inverse;
-}
+  .breadcrumbs__item {
+    @include u-truncate();
+    float: left;
+    max-width: 20rem;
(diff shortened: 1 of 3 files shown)


noahmanger:
Ok, ready for review, whoever has time.

adborden:
:+1: from me. Jen might still want to look at it. I noticed the mocks from https://github.com/18F/openFEC/issues/1595 look like they are dashed `bottom-border` rather than our normal dotted underline, not sure if that was intended?

noahmanger:
I think that's just an artifact of the design tools. (Sketch doesn't let you do dotted). But good eye!

jenniferthibault:
I'll look between usability sessions soon!

jenniferthibault:
Spot on about the underlines: sketch file underlines and CSS underlines will never really look the same. Every underline in a sketch file is a line segment that I have to manually adjust in order to fit the length of the link (which I do a very sloppy job of)

**One quick change:**

On the calendar page, the breadcrumbs bar should be `primary` or federal blue background.

screen shot 2016-04-14 at 5 34 46 pm

**And a question:**
Do we need breadcrumbs on the data landing page and the registration and reporting landing page?

screen shot 2016-04-14 at 5 39 41 pm
screen shot 2016-04-14 at 5 39 55 pm

noahmanger:
Oops. Fixed the color for the calendar breadcrumbs.

I could go either way with having them on the landing pages. I kind of like them as a means of being really obvious about where you are, but I'd be fine dropping them too, since they look a little weird.

jenniferthibault:
Aesthetically speaking I'd prefer we do drop them from the landing pages, but I don't have any data on how useful people would find them there.

noahmanger:
👍 works for me. I removed them.

jenniferthibault:
Great! Looks good and merging.

PR #309, opened 2016-04-13
Fixing truncated text

noahmanger:
Fixes the issue of issue of truncated text in the calendar not showing up on firefox. I replaced the ellipsis with a gradient like so:

image


jenniferthibault:
Visual design 👍 , can't speak for technical review.

PR #310, opened 2016-04-14
Fixes npm version instructions in README

ccostino:
The instructions in our README file are still not correct for actually releasing updates and publishing to npm, and we're either publishing a new version without the changes or failing to push the tags appropriately.

These changes should address the main issue we're having, but I would like to look more closely about the `--tags` option on `git push` and why just the one step doesn't seem to work.  For the time being, two pushes (one for the change to `package.json` itself and another for the tag update) at least ensures that we get exactly what we're expecting.

/cc @noahmanger

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,19 +118,20 @@ We use [Semantic Versioning](http://semver.org/):
 > PATCH version when you make backwards-compatible bug fixes.
 > Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
 
-When merging into master, update the package version and push to npm:
+When changes in master are ready to be released, follow these steps to update the package version and publish to npm:
 
-    npm version patch
-    npm publish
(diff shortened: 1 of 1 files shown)


noahmanger:
Ah! So things were out of order? Thanks for updating.

ccostino:
The order might have been okay but there was definitely that one step missing.  This order makes the most sense to me - update the version via the `npm` tool, which also commits the change for you, get the change and tags up in GitHub, then publish it after all is said and done.

adborden:
You can do a single `git push` if you include either the refspec of the tag i.e. `HEAD` or use `--follow-tags` (introduced in 1.8.3) instead of `--tags`. I think `--follow-tags` is the preferred way.

e.g.
```
git push HEAD --tags
```
or
```
git push --follow-tags
```

To keep it simple, I just always name the refs I'm pushing.

    $ git push v1.8.3

PR #316, opened 2016-04-15
Site menu

noahmanger:
## Summary
Begins the work of introducing the new site nav from https://github.com/18F/openFEC-web-app/issues/1153

In order to reduce duplicate code between the web app and repo, I'm suggesting we have the sub-nav components as handlebars templates in this repo. (Because the menu is only usable with javascript, it shouldn't matter.) 

This makes use of Adobe's [Accessible Mega Menu example](https://github.com/adobe-accessibility/Accessible-Mega-Menu). I just included the file in a `vendor/` directory, since the repo doesn't have a package.json.

## Screenshots
example

## To do
- [x] I still need to figure out the best way to do this on mobile, given a pretty different structure. 
- [x] Write tests


noahmanger:
Ok, I updated the tests here. I'm thinking we should do mobile nav as a separate PR. This has basic nav on mobile, just no sublists:

image

adborden:
> Because the menu is only usable with javascript, it shouldn't matter.

@noahmanger can you clarify? I thought that to meet accessibility requirements, even if javascript is disabled you're okay as long as the nav is still accessible. But if js was required for the nav, wouldn't it no longer be accessible?

I'm new to accessibility, so lay it on me!

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+    !siteNav.$body.hasClass('is-open') &&
+    siteNav.$list.attr('aria-hidden') !== 'false'  &&
+    !siteNav.$toggle.hasClass('active');
+}
+
+describe('SiteNav', function() {
+  before(function() {
+    this.$fixture = $('<div id="fixtures"></div>');
+    $('body').append(this.$fixture);
+  });
```
adborden: I would add an `after` to delete the fixture container when you're done.

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+    this.siteNav = new SiteNav('.js-site-nav');
+  });
+
+  describe('constructor()', function() {
+    it('should set body to jqueryized selector', function() {
+      expect(this.siteNav.$body).to.be.ok;
+      expect(this.siteNav.$body.length).to.be.ok;
+      expect(this.siteNav.$list.length).to.be.ok;
+      expect(this.siteNav.$toggle.length).to.be.ok;
+      expect(this.siteNav.$body.is(this.$fixture.find('nav'))).to.be.true;
```
adborden: I've never seen `is` used this way, neat.

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+    });
+  });
+
+  describe('Desktop configuration', function() {
+    describe('assignAria()', function() {
+      it('should assign arria attributes to the list', function() {
```
adborden: typo "aria"

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+        expect(this.siteNav.$list.attr('aria-label')).to.equal('Site-wide navigation');
+      });
+    });
+
+    describe('initMegaMenu()', function() {
+      it('should append a mega menu to items with [data-submenu]', function() {
+        expect(this.siteNav.$list.find('[data-submenu]').find('.mega')).to.be.ok;
```
adborden: Watch out for this, `find` returns a jquery object, so even if no elements are matched, this will pass.

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+      });
+    });
+  });
+
+  describe('Mobile configuration', function() {
+    beforeEach(function() {
+      $('body').width(400);
+      this.siteNav = new SiteNav('.js-site-nav');
+    });
```
adborden: Consider re-setting the body width in an `after`.

tests/site-nav.js
adborden: I think you can assert on `length` for this.

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+
+    describe('assignAria()', function() {
+      it('should assign arria attributes to the list and toggle', function() {
```
adborden: typo "aria"

tests/site-nav.js
```diff
@@ -0,0 +1,101 @@
...
+        expect(this.siteNav.$list.attr('aria-hidden')).to.equal('true');
+      });
+    });
+
+    describe('toggle()', function() {
+      it('should show and hide the menu', function() {
+        this.siteNav.toggle();
+        expect(isOpen(this.siteNav)).to.be.true;
+        this.siteNav.toggle();
+        expect(isClosed(this.siteNav)).to.be.true;
```
adborden: This is a minor thing, but I prefer to define things close to where they're used, so applying that logic I would move the `isOpen` and `isClosed` functions into the describe block.

noahmanger:
@adborden good question. While accessibility and support for no JS are often tied together, the latter isn't actually a requirement to the former. In the case of a menu, you need to be able to navigate to all of the pages of the website without JS, which is possible through the presence of links on the various pages represented in the top-level nav. Does that make sense?

adborden:
Ah, thank you, that makes sense.

noahmanger:
Ok, just updated the tests @adborden . Let's still hold off on merging this though until we see how it does in usability testing today.

noahmanger:
Talked to @adborden and we want to merge this in and then tackle the mobile implementation next. The current version of the mobile menu is same as previously, except with not child links:

image

@jenniferthibault can we pair on this tomorrow for visual review?

jenniferthibault:
Sure can! A thought—do we want to start previewing the the Legal resources as a coming soon nav tab? I think that would get folks excited, and would be a highly-visible sign of work that's happening.

noahmanger:
Sounds like a good idea!

adborden:
@porta-antiporta Okay to show "Legal resources (coming soon)" as a disabled nav item? https://github.com/18F/fec-style/pull/316#issuecomment-220040465

porta-antiporta:
@adborden yes (though per @edmullen we are also OK with omitting (coming soon) if it's too long of a tab name)

scss/modules/_nav.scss
```diff
@@ -65,13 +38,16 @@
   bottom: 0;
   z-index: $z-navigation;
 
+  &[aria-hidden='true'] {
+    display: block !important;
```
adborden: Maybe add a comment here.

PR #325, opened 2016-04-19
Ie dropdown fix

noahmanger:
This fixes an issue where it wasn't possible to click on an item in a dropdown panel in internet explorer. It appears that what was happening was when clicking on a dropdown option, the target of the event that `handleFocusAway` received was set to the panel itself, rather than the item inside. So I added a step to the conditional to check if `e.target` is the panel. 

I'd love it if someone can check this in both Chrome and IE 11. 

To verify, open a dropdown filter and click on an item. It should work like so:

example

cc @adborden @ccostino


adborden:
I have an IE vm, I'll try to test this out today.

--
Aaron D Borden
Human and Hacker

PR #326, opened 2016-04-19
Restore page header style

noahmanger:
This restores a class that I pulled out in the breadcrumb refactor, but which is still active in the web app for the time being.


ccostino:
Thanks, @noahmanger!

Issue #259, opened 2016-03-11
Add filter tags to new election year filter


noahmanger:

jenniferthibault:
What does this mean @noahmanger ?

noahmanger:
Sorry, this a development task. The election cycle filters on the new candidate office datatable pages need to be configured to have filter tags associated with them.

Issue #312, opened 2016-04-14
To reduce visual tension in the layout, adjust spacing in filter panel template


jenniferthibault:
Some refinements to the filter panels that will make them awwwwwesommmmme.

Desktop only change: **`.filter__header`:**
- Add `margin-top: .6rem`
- Increase `margin-bottom` to 1.5rem from 1rem

![screen shot 2016-04-14 at 10 46 13 am](https://cloud.githubusercontent.com/assets/11636908/14532225/52dd91ee-022e-11e6-9181-dcb55d4ac55c.png)

Existing:

<img width="535" alt="screen shot 2016-04-14 at 10 46 40 am" src="https://cloud.githubusercontent.com/assets/11636908/14532235/6022d1de-022e-11e6-94fb-73380d6f1c09.png">

Desired changes:

<img width="528" alt="screen shot 2016-04-14 at 10 46 28 am" src="https://cloud.githubusercontent.com/assets/11636908/14532244/67a42dea-022e-11e6-8d17-f3089c89fc54.png">

- When there's a search bar immediately under the header, add ~`2.1rem` of margin between the bottom of the header and the top of the search title. This happens on the Candidates and Committees pages. (implemented on all screen sizes):

<img width="650" alt="screen shot 2016-04-14 at 10 52 23 am" src="https://cloud.githubusercontent.com/assets/11636908/14532428/08475d94-022f-11e6-8d2b-c0b0f9ebc630.png">

- When an item is not inside of an accordion, add 1rem margin top between the line and the filter name. (implemented on all screen sizes)

Existing:

- Receipts
<img width="299" alt="screen shot 2016-04-14 at 10 58 10 am" src="https://cloud.githubusercontent.com/assets/11636908/14532651/cd45adbc-022f-11e6-82dc-7fa7a61778ac.png">

- Candidates
<img width="301" alt="screen shot 2016-04-14 at 11 07 24 am" src="https://cloud.githubusercontent.com/assets/11636908/14532989/2e9f6854-0231-11e6-975e-acff8cb7b79f.png">

- Committees
<img width="299" alt="screen shot 2016-04-14 at 11 07 44 am" src="https://cloud.githubusercontent.com/assets/11636908/14532990/2ea078f2-0231-11e6-837b-f3492dc3d226.png">

Desired:
[...]

PR #329, opened 2016-04-21
Adding filter tags to election cycle filters

noahmanger:
## Summary

This adds the ability for the election + cycle filters on candidate office data table pages (e.g. /candidates/president/) to generate filter tags.

Because you always need an election and cycle selected, I made the remove button optional, as there should be no way to remove these tags, only change them with the filters.

I also combined the select and toggles into a single tag, rather than separate ones. Curious what @emileighoutlaw thinks of the microcopy: e.g. `2016 election: 2013-2014`.

## Screenshots

example

## Remaining to dos
- [x] Add tests for the election filter component, which weren't previously there
- [x] Adjust the styling of filters per https://github.com/18F/fec-style/issues/312

## Resolves
Resolves https://github.com/18F/fec-style/issues/259


noahmanger:
Ok! Added tests and ready for review. 

Note: I was having trouble trying to run `hbsfy` and `stringify` together in Karma, so I decided to convert the `feedback.html` template into `feedback.hbs` and remove `stringify`.

jenniferthibault:
The spacing from #312 looks good to me, thanks Noah! After Emileigh looks at the text, you may want @adborden as technical review though, I was only looking at the visuals.

scss/modules/_filters.scss
```diff
@@ -105,16 +113,20 @@
...
+
+.filters__message-container {
+  padding: 0 2rem;
 }
 
 .filter {
   border-bottom: 1px solid darken($neutral, 10%);
-  padding: 1.5rem 0 .5rem 0;
+  padding-bottom: .5rem;
+  margin-top: 1.5rem;
```
adborden: Should be alphabetically ordered, right?

tests/filter-tags.js
```diff
@@ -54,6 +54,12 @@ describe('filter tags', function() {
       var tag = this.tagList.$list.find('li[data-id="name"]');
       expect(tag.length).to.equal(1);
     });
+
+    it('it does not add remove button if nonremovable is true', function(){
+      $(document.body).trigger('filter:added', [{key: 'name', value: 'timmy', nonremovable: true}]);
+      var tag = this.tagList.$list.find('li[data-id="name"]');
+      expect(tag.find('button').length).to.equal(0);
```
adborden: You can also write this as
```javascript
expect(tag.find('button')).to.have.length(0);
```
to give a better error message.

tests/filter.js
```diff
@@ -147,4 +150,77 @@ describe('filter set', function() {
...
+      this.filter.$election.val('2016');
+      this.filter.setTag();
+      expect(trigger).to.have.been.calledWith('filter:added', [
+        {
+          key: 'election',
+          value: '2016 election: 2013-2014',
+          nonremovable: true
+        }
+      ]);
+      $.prototype.trigger.restore();
```
adborden: You should move the spy setup/restore into a `beforeEach`/`afterEach`. IIRC, if the `expect` fails, the test just ends so the cleanup won't actually happen. If it's in an `afterEach`, it will still get cleaned up.

scss/modules/_filters.scss
noahmanger: 👍

tests/filter-tags.js
noahmanger: 👍

noahmanger:
Thanks for the feedback @adborden . Just pushed the changes.

emileighoutlaw:
Looking at the microcopy this second. Can we hold a merge for 30 minutes until I thumbs up?

emileighoutlaw:
I am struggling so hard to read this JS.

But having a single tag, a lá `2016 election: 2013-2014` is a great idea, and if that's what the text output looks like, 👍  from me

noahmanger:
Cool! Yep, that's what it is.

PR #341, opened 2016-04-29
Fixes glossary definition

emileighoutlaw:
Content only: I copied the wrong definition into the glossary under political party. This is the correct one. 

cc @ccostino


ccostino:
Thanks, @emileighoutlaw!

PR #342, opened 2016-05-02
Add a few subclasses for doc legal search

adborden:
## Summary

Adds a few styles needed for the legal document search results.

## Screenshots

`.filters--fixed`
screenshot from 2016-05-02 11-58-30

`.results-info--bottom`
screenshot from 2016-05-02 12-00-03

### DIFF
--- a/scss/modules/_datatables.scss
+++ b/scss/modules/_datatables.scss
@@ -1,5 +1,41 @@
 // Datatables
 //
+// Standard design for tabular data. If the table is
+// preceded/followed by buttons (or nav arrows), always enclose with
+// a rule line using `.results-info`.
+//
+// Markup:
+// <table id="data-table" class="data-table {{ modifier_class }}">
+//   <thead>
+//     <tr>
+//       <th>No.</th>
+//       <th>Name</th>
+//       <th>Score</th>
+//     </tr>
+//   </thead>
+//   <tbody>
+//     <tr>
+//       <td>1</td>
+//       <td>Red</td>
+//       <td>0.1234</td>
+//     </tr>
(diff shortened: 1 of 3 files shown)


noahmanger:
Code looks fine, but a pattern question for @jenniferthibault / @edmullen : 

In other instances ([like here](https://beta.fec.gov/data/committee/C00213512/?tab=receipts)), we have a rule above and below the table controls at the bottom (`.results-info--bottom`). Do we want to keep that consistent? Or is the distinguishing factor that the this instance has the note below?

image

jenniferthibault:
When building that out the first time around, the distinguishing factor was the pagination buttons, (rather than the note text) so I would like to add the line underneath here please.

adborden:
@noahmanger this is good to merge? I'll make @jenniferthibault 's change in the web-app.

noahmanger:
Tiny thing, but if we're not using the `.results-info--bottom` class can we just remove it here?

adborden:
Yeah, just pinged @jenniferthibault to clarify and yes, we won't be needing this style anymore. I removed the `.results-info--top` as well and added some documentation from Jennifer.

screenshot from 2016-05-17 21-34-39

noahmanger:
Rad. Just because I don't want to forget, are you replacing the couple instances of `.results-info--top` in the web app?

adborden:
Technically this is a major version bump, because it removes an existing style and would potentially break things. I made the change in the web app, so make sure to merge that before bumping the fec-style version there. https://github.com/18F/openFEC-web-app/issues/1213.

I did not see any uses of `.results-info--top` in the cms.

Issue #311, opened 2016-04-14
Typeset note text on profile pages


jenniferthibault:
Readability could be improved with more leading and tighter tracking:

<img width="1203" alt="screen shot 2016-04-13 at 3 51 15 pm" src="https://cloud.githubusercontent.com/assets/11636908/14531340/f06f94d8-022a-11e6-8e64-b2a6b35a8027.png">

jenniferthibault:
I changed strategies a bit, and instead of adjusting the tracking and leading, changed the font style to use  Gandhi Italic, which is both more legible in paragraph form, and will be used consistently as a note style. This will be resolved when https://github.com/18F/fec-style/pull/345 and https://github.com/18F/openFEC-web-app/pull/1210 are merged.

PR #345, opened 2016-05-09
Typesetting note styles

jenniferthibault:
## Summary

Restyling note text from Karla (sans serif) to Gandhi italic (serif)  to address legibility issues in https://github.com/18F/fec-style/issues/311 and make the style more usable for different settings. 

## Screenshots

Something weird for me is that the change isn't showing up in the style guide, but it IS showing up in the web app that's running off the style guide. Example:

**Style guide**
screen shot 2016-05-09 at 12 25 24 pm

**In the wild, after**
screen shot 2016-05-09 at 11 21 07 am
screen shot 2016-05-09 at 11 21 21 am

**Before shots**
screen shot 2016-05-09 at 11 22 35 am
screen shot 2016-05-09 at 11 22 51 am

cc @noahmanger , @adborden


adborden:
@jenniferthibault yeah, I've noticed the styleguide not getting updated on build either. I think this is related https://github.com/18F/fec-style/pull/346

adborden:
:+1: 

@noahmanger why is the kss.css committed to the repo? Isn't it a generated file that gets created on build? Should it be included in this PR?

jenniferthibault:
Actually, hold off on merging this in, I'm going to set up a corresponding PR to the CMS.

noahmanger:
Oh yep, that can be ignored.

scss/components/_list-styles.scss
```diff
@@ -21,6 +21,9 @@
     margin-left: 2rem;
     padding: .5rem 0;
     font-size: 1.6rem;
+
+    &:last-child {
+     margin-bottom: 1.5rem;
```
ccostino: Is there a closing curly brace missing here?

ccostino:
Speaking of `fec-template/public/kss.css`, shouldn't that just be in the `.gitignore` file? Is there a reason why it's not?

jenniferthibault:
Ok, I think I've come about as far as I can go without someone who actually know what they're doing 😄 

## Dependent PR's:
- CMS: https://github.com/18F/fec-cms/pull/295
- Web app: https://github.com/18F/fec-style/issues/311

## Question
- How do I ignore the kss.css?

## Changes:
- Restyles `h4` for Registration & Reporting side. This style is seldom used in our current text structures, but if it starts showing up more, is built to work nicely as a transition between `h3` and `h5`, which is 1.6rem Gandhi Serif bold — so the same size as body copy, but at a header level. 
  - Warning: I want to watch out for times when h4/h5 styles are implemented in Wagtail, because it looks like instead of proper headers, sometimes just `b` is applied to body copy, and sometimes extra line breaks are applied which are wreaking some havoc on optical margin top/bottoms.  
- Builds in space between the end of paragraph text and headline text 
- Builds in space after the end of a bulleted list

screen shot 2016-05-09 at 6 24 16 pm

@noahmanger I would love your eyes on the adjacent sibling selectors in context of the CMS pages: I don't think I have the template for article pages built out on Wagtail, because I keep getting 404's locally, so I'm shooting from the hip here.

scss/components/_list-styles.scss
```diff
@@ -21,6 +21,10 @@
     margin-left: 2rem;
     padding: .5rem 0;
     font-size: 1.6rem;
+
+     &:last-child {
```
jenniferthibault: OK HOUND. I don't know what you want from me.

PR #340, opened 2016-04-29
New checkboxes

noahmanger:
## Summary
New checkbox styles based on https://github.com/18F/fec-style/issues/314

Resolves https://github.com/18F/fec-style/issues/320

## Screenshots
image

_cc @username_

PR #346, opened 2016-05-09
Remove `styleguide` dir on build

adborden:
Fixes typo where `styleguide` directory is not removed on build.


noahmanger:
Thanks for catching this. I noticed that since it was actually removing the directory, we needed to also include `copy-images` at the end. I just included this in https://github.com/18F/fec-style/pull/340 so we can just merge it all there.

noahmanger:
And with that, I'll just close this.

PR #348, opened 2016-05-16
Padding on terms

noahmanger:
## Summary
Small change to add padding between items in the detail panels.

Should be merged with https://github.com/18F/openFEC-web-app/pull/1211

cc @ccostino


ccostino:
When https://github.com/18F/openFEC-web-app/pull/1211 is merged I'll merge this right after.  I saw a bit of extra discussion at the end when I went to merge it earlier so I'm holding off for a moment.

PR #349, opened 2016-05-18
Improved typeahead interaction

noahmanger:
## Summary
Includes numerous changes made to improve the usability of the typeahead filters:
- Adds helper text to free-text options: "Search for:" / "Select a committee:"
- Disables the search button by default and enables it only after you can search for something
- Selects the first option both on enter and clicking the search button
- Adds a more noticeable highlight state on whichever option will be submitted

Also, I'm sure there are more changes to make to this component, but I'd prefer to get this merged in sooner than later (once tests are updated). But before I update tests, @ccostino or @adborden can you take a look?

## Screenshots
image

## Notes:
@onezerojeremy and @jenniferthibault : did we decide based on usability testing if we want to keep the down arrow vs. the search icon in the button? 

## To do:
- [x] Update unit tests


jenniferthibault:
@noahmanger : per Jeremy's notes in [the usability testing issue](https://github.com/18F/openFEC/issues/1594#issuecomment-219451518)
> the dropdown indicator was not universally understood as a way to find an item by searching through a closed list of items, consider a new icon that is neither the magnifying glass search icon or the drop down list icon (though if choosing between the two, it seems the magnifying search indicator was a better trigger).

I haven't had a chance to visit this issue yet, but planning on it today

ccostino:
@noahmanger, so far this all looks okay to me.  Nothing is jumping out at me as problematic and I just ran things locally and they seem to be in good order.  I say get the tests in and we'll give it a final review to get it merged. :-) Thanks!

noahmanger:
Ok, I updated unit tests @ccostino . There were basically no tests before (I recall this being intentional because we were still tweaking the design), so I didn't get _everything_ covered, but I got the big functions and my new ones covered.

I'm never sure if I'm writing tests in a smart way. They pass, but if folks have feedback on better ways to test things, please share.

ccostino:
@noahmanger, nicely done!  The tests are concise and focus on small, actionable things and I think you cover a lot of ground here so far.  @adborden, I am going to defer to your judgment if you have a moment to take a look as I think you're probably more versed in this area than I am.  However, I like what I see and think you're on the right track with things, @noahmanger!

adborden:
I notice that if you arrow down off the list, there's no highlight on the first item. Arrow down once more, and you're back on the first item. That one step that doesn't have a highlight is confusing.

tests/typeahead-filter.js
```diff
@@ -29,18 +29,130 @@ describe('typeaheadFilter', function() {
...
+    this.typeaheadFilter.handleAutocomplete({}, {id: '12345'});
+    expect(this.typeaheadFilter.datum).to.deep.equal({id: '12345'});
+  });
+
+  it('should submit on enter', function() {
+    var handleSubmit = sinon.spy(this.typeaheadFilter, 'handleSubmit');
+    this.typeaheadFilter.handleKeypress({keyCode: 13});
+    expect(handleSubmit).to.have.been.calledWith({keyCode: 13});
+    this.typeaheadFilter.handleSubmit.restore();
+  });
```
adborden: This is fine to write this way, but I prefer splitting the spy stuff into `beforeEach`/`afterEach`. I try to keep `it` blocks to only contain an action and the assertion. The rest is setup/cleanup.

tests/typeahead-filter.js
```diff
@@ -29,18 +29,130 @@ describe('typeaheadFilter', function() {
...
+    this.typeaheadFilter.enableButton.restore();
+    this.typeaheadFilter.disableButton.restore();
+  });
+
+  it('should clear input', function() {
+    this.typeaheadFilter.$field.val('hello');
+    this.typeaheadFilter.enableButton();
+    this.typeaheadFilter.clearInput();
+    expect(this.typeaheadFilter.$field.val()).to.equal('');
+    expect(this.typeaheadFilter.$button.hasClass('is-disabled')).to.be.true;
```
adborden: To get a more descriptive error, you could write this with a string `match`. e.g.

```javascript
expect(this.typeaheadFilter.$button.class()).to.match(/\bis-disabled\b/);
```

Which would give the error message,
> expected "button button-thing" to match "\bis-disabled\b" but it did not.

If you're not familiar, `\b` matches a word boundary.

adborden:
Tests and everything look great!

noahmanger:
Thanks!

PR #351, opened 2016-05-18
Calendar visual tweaks

noahmanger:
## Summary

- Styles the list view based on https://github.com/18F/fec-cms/issues/298
- Sets default view to list view
- Hides text on view toggles
- Wraps filters in checkboxes so that the lines break correctly

To do:
- [ ] Add tooltip text to tags (or link to glossary terms?)

## Screenshots
image

cc @jenniferthibault

### DIFF
--- a/scss/modules/_calendar.scss
+++ b/scss/modules/_calendar.scss
@@ -500,6 +500,13 @@
     .fc-button-group {
       float: right;
     }
+
+    .button {
+      background-position: 50% 50%;
+      padding-left: 2rem;
+      text-indent: -9999px;
+      width: 4rem;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


noahmanger:
@jenniferthibault and @emileighoutlaw how do you feel about merging this in without the tooltips / glossary links while we sort that out? Do we want to include this in the release?

jenniferthibault:
I don't think it's high risk to include, and since it's an FEC priority to improve the calendar issues, (and this does) I'm ok with it going into the release without the tooltips, and talking through the changes with the whole team on Monday.

emileighoutlaw:
Been merged already, but 👍 from me, too

PR #352, opened 2016-05-18
Fixing a couple weird things with buttons

noahmanger:
## Summary
Corrects for a few things that slipped through in the last few patches, such as new borders put on buttons that shouldn't be there and incorrect padding on the search button.


adborden:
Sweet.

PR #358, opened 2016-05-23
Release fixes

noahmanger:
## Summary
- Menu improvements:
  - Content changes
  - Adding typeahead to candidate search
  - Removing mega menu from DOM on window resize
  - Updates link to individual contributions page to include `is_individual` and `max_date` filters
- Fixing broken icons in IE
  - Adds `charset=UTF8` to the SVG strings generated by the Gulp plugin (required for IE)
  - Removes the Gulp step that removed dimensions from SVGs — IE requires these to be set in order scale the icons using `background-size`
- Adjusts spacing on messages in filter panel


gulpfile.js
```diff
@@ -11,9 +11,6 @@ gulp.task('minify-icons', function() {
     .pipe(svgmin({
       plugins: [
         {
-          removeDimensions: true
```
adborden: By removing this, do we still need to set the `background-size` on a per-icon basis?

js/site-nav.js
```diff
@@ -65,6 +67,14 @@ SiteNav.prototype.initMegaMenu = function() {
       focusClass: 'is-focus',
       openClass: 'is-open'
     });
+
+    new typeahead.Typeahead('.js-menu-search', 'candidates', '/data/');
+  }
+};
+
+SiteNav.prototype.destroyMegaMenu = function() {
+  if ( $('body').width() < helpers.BREAKPOINTS.LARGE ) {
```
adborden: Will the menu come back on it's own if you expand the window back above the breakpoint?

js/templates/nav-data.hbs
```diff
@@ -65,8 +70,8 @@
         </div>
         <ul>
           <li class="mega__item"><a href="{{webAppUrl}}/filings">All filings</a></li>
-          <li class="mega__item is-disabled">Authorized committees</li>
-          <li class="mega__item is-disabled">Unauthorized committees</li>
+          <li class="mega__item is-disabled">Authorized committee reports</li>
+          <li class="mega__item is-disabled">Unauthorized committees reports</li>
```
adborden: Is the copy intentionally use different pluralization? Authorized singular committee vs Unauthorized plural committees?

noahmanger: Oops, that was a mistake. Thanks!

js/site-nav.js
noahmanger: Nope. The library we're using doesn't have a `destroy()` method, so bringing it back re-initailizes it all together and rebinds all the event handlers and everything. I think it's ok though. It's more of a temporary fix until we get the mobile version going.

gulpfile.js
noahmanger: Not positive, but I think it might. I want to wait until after launch to mess with it though.

PR #359, opened 2016-05-24
Sticky bar

noahmanger:
## Summary

Adds a new JS component and related styles for sticky bars for use on candidate, committee and election pages (for the tabs), and on data pages (for the tag area). 

This is different than the other Sticky npm module we're using, which is used for the side-nav / table of contents on certain pages. That library is more complicated and works well for that task, but for reasons I was unable to figure out, couldn't handle this use case well. Ultimately it was just faster to write my own component to replace it for this scenario.

## Screenshots
demo

### DIFF
--- a/js/sticky-bar.js
+++ b/js/sticky-bar.js
@@ -1,0 +1,27 @@
+'use strict';
+
+var $ = require('jquery');
+
+function StickyBar(selector) {
+  this.$body = $('body');
+  this.$bar = $(selector);
+  this.height = this.$bar.outerHeight();
+  this.offset = this.$bar.offset().top;
+  this.delay = 100; // Delay before it sticks
+
+  this.defaultBodyPadding = this.$body.css('padding-top');
+  $(window).on('scroll', this.toggle.bind(this));
+}
+
+StickyBar.prototype.toggle = function() {
+  var scrollTop = this.$body.scrollTop();
+  if (scrollTop >= this.offset + this.delay) {
+    this.$bar.addClass('is-stuck');
+    this.$body.css('padding-top', this.height);
+  } else if (scrollTop < this.offset) {
+    this.$bar.removeClass('is-stuck');
+    this.$body.css('padding-top', this.defaultBodyPadding);
+  }
+};
+
+module.exports = {StickyBar: StickyBar};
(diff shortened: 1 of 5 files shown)


jenniferthibault:
@noahmanger is it possible to make the bar stick the moment the top of it hits the top of the screen? It's a little disorienting to me that the bar disappears for a moment, then reappears once you hit the bottom of it.

noahmanger:
It is, though I was finding that that too was a little disorienting, as there was a slight jitter. I found this to be smoother, though we can try it both ways.

noahmanger:
@jenniferthibault let's pair on this today so you can help me figure out where the right point is.

js/sticky-bar.js
```diff
@@ -0,0 +1,27 @@
+'use strict';
+
+var $ = require('jquery');
+
+function StickyBar(selector) {
+  this.$body = $('body');
+  this.$bar = $(selector);
+  this.height = this.$bar.outerHeight();
+  this.offset = this.$bar.offset().top;
+  this.delay = 100; // Delay before it sticks
```
adborden: `delay` makes me think time, but here you're actually using it as a distance offset. Unfortunately `offset` is already taken, but I would consider renaming this. Maybe `triggerOffset`?

adborden:
Curious what you two work, out, but I wonder why the need for the animation at all? I think if it just sticks once you hit the top of the element, it looks pretty smooth, no?

noahmanger:
In some cases, like on the datatable pages, you're right, it just sticks at the top, so we decided to not have it transition or have a delay. But in others, like on election / committee / candidate pages, we show a new row in the sticky header when it's stuck, which makes for a jarring experience if there's no transition. Here's how it looks with the transition:
ex

But the transition is now optional. Applying the `with-transition` class adds the CSS transition and adding `data-trigger-offset` sets an additional offset on the trigger so it waits until it's fully offscreen before sticking it.

jenniferthibault:
Thanks for pairing! The interaction looks good on my end.

scss/components/_stickybar.scss
```diff
@@ -8,16 +8,24 @@
...
     top: 0;
     left: 0;
   }
+
+  &.with-transition {
+    @include transition(top .4s ease-out);
+    top: -200px; // Setting this property so we can animate it when .is-stuck is applied
+
+    &.is-stuck {
+      @include transition(top .4s ease-out);
```
adborden: I don't think you need this, since the `with-transition` class already has the transition property.

PR #361, opened 2016-05-24
Adding support for changing typeahead suggestion

noahmanger:
## Summary

Creates a simple version of a new `FilterControl` component which can be set to trigger `filter:modify` events that change the settings of a particular filter. The only instance of this currently is for changing the way the `contributor name` filter shows typeahead suggestions depending on if the filter above has only `individual` checked or not. 

As such, the code is not fully abstracted, but I think that's ok for now, and we can abstract later. 

Can you take a look @ccostino or @adborden ? If this looks like a good approach, I'll go ahead and add tests.

## Screenshots
image

### DIFF
--- a/js/filter-control.js
+++ b/js/filter-control.js
@@ -3,30 +3,28 @@
 var $ = require('jquery');
 
 function FilterControl($selector) {
-  this.$body = $selector;
-  this.modifiesFilter = this.$body.data('modifies-filter');
-  this.modifiesProperty = this.$body.data('modifies-property');
-  this.getValue();
+  this.$element = $selector;
+  this.modifiesFilter = this.$element.data('modifies-filter');
+  this.modifiesProperty = this.$element.data('modifies-property');
 
-  this.controlledFilter = $('#' + this.$body.data('controls'));
-  this.$body.on('change', this.handleChange.bind(this));
+  this.controlledFilter = $('#' + this.$element.data('controls'));
+  this.$element.on('change', this.handleChange.bind(this));
 }
 
 FilterControl.prototype.getValue = function() {
   var values = [];
-  this.$body.find('input:checked').each(function() {
+  this.$element.find('input:checked').each(function() {
     values.push($(this).val());
   });
-  this.values = values;
+  return values;
 };
 
 FilterControl.prototype.handleChange = function() {
-  this.getValue();
-  this.$body.trigger('filter:modify', [
+  this.$element.trigger('filter:modify', [
     {
       filterName: this.modifiesFilter,
       filterProperty: this.modifiesProperty,
-      filterValue: this.values
+      filterValue: this.getValue()
     }
   ]);
 };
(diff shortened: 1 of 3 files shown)


js/typeahead-filter.js
```diff
@@ -167,43 +174,67 @@ TypeaheadFilter.prototype.appendCheckbox = function(opts) {
...
+
+TypeaheadFilter.prototype.changeDataset = function(e, opts) {
+  // Method for changing the typeahead suggestion on the "contributor name" filter
+  // when the "individual" or "committee" checkbox filter is changed
+  // If the modify event names this filter, destroy it
+  if (opts.filterName === this.fieldName) {
+    this.$field.typeahead('destroy');
+    // If the value array is only individuals and not committees
+    // set the dataset to empty and re-init
+    if (opts.filterValue.indexOf('individual') > -1 && opts.filterValue.indexOf('committee') < 0) {
```
ccostino: If we were to abstract this out further for a more general use-case I'd be worried about accounting for an arbitrary number of checkboxes here.  We'd need another way to manage that in a more scalable and flexible manner.

ccostino:
I only had one major concern when looking this over, and I know that's more of a future state thing but figured I'd call it out.  Otherwise it looks like a solid approach to me, thanks @noahmanger!

js/typeahead-filter.js
noahmanger: Agreed. Yeah, this was particularly brittle right here, but I think I'd rather abstract later when we know what we're abstracting for.

ccostino: Yes, I definitely agree on that point!

js/filter-control.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+var $ = require('jquery');
+
+function FilterControl($selector) {
+  this.$body = $selector;
```
adborden: Not a big deal, but I keep seeing `$body` and thinking the entire `body` element. I know `$body` is the convention, but I usually use the name `$element` for this kind of stuff. Maybe we can change this going forward?

js/filter-control.js
```diff
@@ -0,0 +1,34 @@
...
+  this.$body.on('change', this.handleChange.bind(this));
+}
+
+FilterControl.prototype.getValue = function() {
+  var values = [];
+  this.$body.find('input:checked').each(function() {
+    values.push($(this).val());
+  });
+  this.values = values;
+};
```
adborden: Rather than keeping `values` as state on the the `FilterControl`, you can just return `values`. Since you're using event triggering anyway to pass the values, it doesn't need to be kept as state. I prefer to avoid state and state mutation when possible. It avoids a whole class of bugs and debugging headaches, IMHO.

js/filter-control.js
noahmanger: +1. I'm not sure why the convention started, but I like `$element` more.

jenniferthibault:
👍  on approach and design. Is this waiting for any more review from you @adborden ?

Issue #286, opened 2016-03-29
Global font size makes it difficult to re-use components with other frameworks


adborden:
FEC style defines a reduced [62.5% font size](https://github.com/18F/fec-style/blob/master/scss/elements/_typography.scss#L15) and then uses a lot of `rem` units for padding, margin, height, font-size in components like the disclaimer and masthead.

When integrating the FEC styles with [eregs](https://github.com/18F/fec-eregs), the global font size is 100%, which really throws off some of the component sizes using `rem`s. In this particular case, reducing the global font-size has too many undesired consequences.

I'm overriding the styles of the FEC components using fixed sizes (px), but I wonder if using a different unit (`em`) for some of these components would be more appropriate, since things like the masthead and disclaimer are probably designed to be a very particular size, no matter what the rest of the site is scaled to.

Wondering if you folks have any thoughts on this kind of problem.

adborden:
The [Frontend guild's CSS styleguide](https://pages.18f.gov/frontend/css-coding-styleguide/units/) suggests setting the global font-size to 10px.

noahmanger:
Good points! That should be fine. Just asking in #g-frontend to see what others think.

noahmanger:
I opened up an issue for discussion in the front-end guide: https://github.com/18F/frontend/issues/118

noahmanger:
So, the conversation over at the [web design standards on this issue](https://github.com/18F/web-design-standards/issues/992) is coming down on the side of setting the base font-size to a percentage. 

As for whether that percentage is 100% or 62.5% only matters in as much as it affects a given situation (like this one). 

I'm not opposed to changing fec-style to 100% base font-size, but then it would just entail updating probably hundreds of REM declarations to use the new math. I'm not familiar with eregs or what you're trying to do there, so maybe there's another solution?

noahmanger:
Is there anything further you want to do on this, or can we icebox until we decide which method we're going to use for eregs?

adborden:
@noahmanger I'm following up on this now as part of looking at how to integrate `fec-style` with `fec-eregs`. Looking at some other frameworks to see what strategies they're using. I would leave this open for a little longer.

adborden:
Okay, finally spent some time this sprint on this problem.

tl;dr, I think instead of using fixed sizes everywhere, we should use a function which calculates the size for us. e.g.

```scss
$base-size-scale-factor: 10; // This value would be different for betaFEC than for eregs
@function u($size) {
[...]

PR #369, opened 2016-05-25
Feature/base size

adborden:
## Summary

In order to use FEC's styleguide with other frameworks, like eRegs, it needs to be configurable to size components properly. All of FEC's components are based on `rem`s so the components are sized based on the browser's default font size. This works well, except when combined with eRegs, where the browser's effective `rem` size is larger. This ends up with FEC components looking gigantic.

This change introduces a `u()` function used to scale `rem`-based sizes at build time. Two variables `$base-size-px` and `$px-only` affect how the sizes are calculated. For FEC based projects, this means nothing, except that instead of writing `2rem`, you would write `u(2rem)`. For eRegs, by setting these two variables, the FEC components are scaled down and play nice with eRegs.

Note, I also moved the documentation for Utility Mixins to the bottom of the styleguide table of contents. These are advanced sections, not required for understanding the FEC design system, so they seem more appropriately described at the end.

The conversion of the units was done with this (GNU) sed script:
```
# sed -r -i -f wrap-rem-units.sed

# A comment, but jump to `a` to process it
\:rem.*//: b a

# A comment, skip the line.
\://: { p ; d }

: a 

# Wrap the line with the function. i.e. u(1rem 0 10px 0); Matches include:
# font-size: 1rem;
# padding-right: -1rem;
# margin-right: -.1rem;
# margin-right: -12.1rem;
# margin: 1rem 0 0 0;
# margin: 1rem 0 10px 0;
/rem/ s/:\s*((-?([0-9]+)?(\.?[0-9]+)(\w+)?\s*)+)\s*;/: u(\1);/
```

## Screenshots

Creates a new section documenting the new functions.
screenshot from 2016-05-25 15-40-45

cc @noahmanger @msecret @anthonygarvan @annalee

### DIFF
--- a/scss/_functions.scss
+++ b/scss/_functions.scss
@@ -1,0 +1,80 @@
+// Utility Functions
+//
+// Various useful functions.
+//
+// Weight: 90
+//
+// Styleguide functions
+
+// Unit conversion
+//
+// When using the FEC styleguide with other frameworks, it might be useful to
+// scale and convert the FEC component sizes. Use these variables to adjust how
+// the styleguide is built.
+//
+// $px-only - (bool) convert `rem` units to `px`
+// $base-size-px - (integer) sets the base size for a `rem` unit (only take effect when `$px-only` is true)
+//
+// Styleguide functions.convert-units
+$base-size-px: 10 !default;
+$px-only: false !default;
+
+// u-parse-int($n)
+//
+// Parses the integer part from a value with unit.
+//
+// $n - the value to parse e.g. 2em
+//
+// Styleguide functions.parse-int
+@function u-parse-int($n) {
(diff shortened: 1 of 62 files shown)


adborden:
Fixes https://github.com/18F/fec-style/issues/286

noahmanger:
This looks great! Totally makes sense. Good call moving utilities to the bottom. Let me know when you've had a chance to fix the conflicts. Might be good to pull down master and run the script to account for anything new merged in.

It'd be great if we could add a plugin to scss-lint to make sure we always use the `u()` function...but this is maybe a topic for another day.

You should totally share this out with the frontend group when you're done, given how the topic has come up in other places.

adborden:
Rebased.

PR #372, opened 2016-05-26
Mobile nav

noahmanger:
## Summary

Includes styles, markup and interaction for new mobile menu. The top-level nav (e.g. the first panel when you first open it) is just a responsive version of the primary site-nav bar:

image

Given the very different markup needs for the mega menu vs the mobile menu, I created an entirely separate handlebars template that is then appended to the DOM at certain breakpoints (and removed at others). This file consists of various `.site-nav__panel--sub` divs, each of which is triggered by a button (and then closed by a corresponding back button, which is labeled for the previous section).

## Screenshots
demo

This is a little tricky and the code could certainly be cleaned up, so would appreciate extra eyes. 

## Remaining tasks:
- [x] Clean up Sass code
- [x] Double-check all links
- [x] Update style guide documentation
- [x] Add tests for JS


noahmanger:
Ok this is ready for review.

tests/site-nav.js
```diff
@@ -75,33 +76,74 @@ describe('SiteNav', function() {
...
       it('should assign aria attributes to the list and toggle', function() {
         expect(this.siteNav.$toggle.length).to.be.ok;
         expect(this.siteNav.$toggle.attr('aria-haspopup')).to.equal('true');
-        expect(this.siteNav.$list.attr('aria-hidden')).to.equal('true');
+        expect(this.siteNav.$menu.attr('aria-hidden')).to.equal('true');
       });
     });
 
     describe('toggle()', function() {
       function isOpen(siteNav) {
```
adborden: Another strategy for this, you can name your function like `assertIsOpen()` and instead of returning a boolean, it can just throw an assertion error to fail the test. You could use `expect()` statements in here too which would improve the messaging in a test failure.

tests/site-nav.js
```diff
@@ -75,33 +76,74 @@ describe('SiteNav', function() {
...
+
+  describe('switchMenu()', function() {
+    before(function() {
+      this.siteNav = new SiteNav('.js-site-nav');
+      $('body').width(400);
+    });
+
+    afterEach(function() {
+      $('body').width(1000);
+    });
```
adborden: You have an imbalanced before/after, beforeEach/afterEach which on fist glance looks like a bug. I would split this up into two sub-`describe`s, one testing going from small to big, the other from big to small.

```javascript
describe('switchMenu()', function () {
  describe('when resizing small to large screen', function () {
    before(function() {
      // set mobile screen size
      // init .js-site-nav
      // assert mobile menu exists for sanity
      // set screen size large
    });
    after(function () {
      // restore to original size
    });
    it('has mega menu', function() {
      // assert mobile menu is removed
      // assert mega menu is in correct state
    });
  });

  describe('when resizing large to small screen', function () {
    // Similar to above
  });
});

File: tests/site-nav.js [hunk 3]
```diff
@@ -75,33 +76,74 @@ describe('SiteNav', function() {
...
+
+    it('should remove the mega menu if the screen gets small', function() {
+      this.siteNav.switchMenu();
+      expect(this.siteNav.$body.find('.mega').length).to.equal(0);
```
adborden: We can also assert the the mobile menu exists, right?

tests/site-nav.js
```diff
@@ -75,33 +76,74 @@ describe('SiteNav', function() {
...
+    });
+
+    it('should remove the mobile menu if the screen gets big', function() {
+      this.siteNav.switchMenu();
+      expect(this.siteNav.$body.find('.js-mobile-nav').length).to.equal(0);
```
adborden: Can we assert that the mega menu is in the correct state, or simply exists?

jenniferthibault:
Successfully got this running locally to check it out 💪 

I added one commit to add space around the back button, but please take a look to make sure I did that with the right approach:

screen shot 2016-06-02 at 10 41 19 am


After navigating around a bit, there's one change I'd like to make. The deeper you go, it becomes hard to tell when a section is also a link to its own page. I think it would be helpful if we added back in underlines to pages that can also be tapped to navigate to them.  
For example:

screen shot 2016-06-02 at 10 37 00 am

And a change that I won't call required, but an element of the original design was to keep the icons. Are you interested in adding them into the same places we have them on the desktop menu? Or if you have thoughts on why it's better to leave them out, i'd be interested in hearing!

screen shot 2016-06-02 at 10 43 26 am

noahmanger:
@jenniferthibault how's this? 
image

image

scss/modules/_nav.scss
```diff
@@ -143,10 +143,14 @@
...
 
+.site-nav__link--underline {
+  border-bottom: 1px dotted $inverse;
+}
+
 .site-nav__back {
   background-position: 0 50%;
   border: none;
-  margin: u(1rem 0 1rem 0)
+  margin: u(1rem 0 1rem 0);
```
jenniferthibault: good catch, thank you!

jenniferthibault:
💣 💥  🚢

Issue #375, opened 2016-05-31
Add style for alt--primary dropdown lists


jenniferthibault:
Just caught this [on the calendar](https://beta.fec.gov/calendar/), which looks to be a misapplied style, or applying a style that we don't have set?

![screen shot 2016-05-31 at 9 23 15 am](https://cloud.githubusercontent.com/assets/11636908/15687963/b7d75de6-2745-11e6-9856-28390f44ba25.png)

I believe that the calendar is the first place we've implemented a dropdown style that isn't the default neutral. Looking at [the style guide](https://pages.18f.gov/fec-style/section-components.html#kssref-components-dropdowns), I don't believe we have a style set for this. Is that correct @noahmanger ? Or do we have one, but haven't listed it in the style guide? 

Since this makes the feature illegible, glad to prioritize it for this week's release, once I know which scenario we're in.

noahmanger:
Got it! Simple fix.

PR #377, opened 2016-05-31
Setting color on dropdown values

noahmanger:
## Summary

Sets the color of `dropdown__value` items, so that when they're in a container that sets font color to something else, they always stay `$base`. 

Resolves https://github.com/18F/fec-style/issues/375

## Screenshots
image

### DIFF
--- a/scss/components/_dropdowns.scss
+++ b/scss/components/_dropdowns.scss
@@ -53,10 +53,6 @@
   }
 }
 
-.dropdown__value {
-  margin: 0;
-}
-
 .dropdown__panel {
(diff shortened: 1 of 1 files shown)


jenniferthibault:
Interesting that's why it was happened! Wouldn't have thought it came from the other container. Nice fix 😄

noahmanger:
Yeah in general it's nice to have colors defined by parent elements, so that you don't have to go through and change individual elements for each situation; but in this case, because the dropdown panels are always white background, we should have set the color on the elements rather than inheriting from parents.

PR #380, opened 2016-06-02
Adding select filter for transaction peroid

noahmanger:
## Summary
Adds and updates components to accommodate the new `transaction_period` parameter for partitioned schedule tables, which necessarily modifies the beginning and ending receipt / disbursement date. Here's how the interaction works. 

1. When you first visit the page with no query parameters (e.g. just `/receipts`), it defaults to the current year, and automatically sets the min and max dates in that two-year period:
image

2. This is reflected in the tags:
image

3. You can edit the date fields to narrow the range, but if you go beyond the range, it shows an error:
image

4. If you fix your mistake, the error goes away.

5. If you change the select, the min and max dates change:
gif 

How does this interaction look to folks? It's definitely a departure from anything we've done so far, but it feels pretty natural to me. cc @jenniferthibault @onezerojeremy 

## Remaining tasks:
- ~~[ ] Update tests~~ UPDATE: Will do after design refinements
- [x] Get proper language in there for the labels and error message (cc @emileighoutlaw )
- [x] Don't load table results if the field doesn't pass validation
- [x] Fix the bug where you can't hit the back button

Goes with a related PR on the web app.


emileighoutlaw:
I'm digging the labels and the interaction. For the alert (when you enter a date outside the transaction period), I like this content:

> You entered a receipt date that's outside your selected transaction period. Please enter a receipt date from **2015–2016**.

jenniferthibault:
The interaction seems good a first test to me! Love that we are using the interface to help people enter the right data, then offering them suggestions to fix it when it goes wrong.

noahmanger:
Ok this is ready for review alongside https://github.com/18F/openFEC-web-app/pull/1242

I'd like to make sure that this new filter code is good at the same time that we're putting the finishing touches on the API. 

CC @ccostino

ccostino:
D'oh, failing tests:

```
PhantomJS 1.9.8 (Linux 0.0.0) filter set date range filters pulls values from query FAILED
	AssertionError: expected '02-01-2050' to equal '2015-01-01'
	    at /tmp/5bf85d1fd7519a4a795c688df1c47749.browserify:2698 <- node_modules/chai/lib/chai/assertion.js:111:0
	    at assertEqual (/tmp/5bf85d1fd7519a4a795c688df1c47749.browserify:3270 <- node_modules/chai/lib/chai/core/assertions.js:493:0)
	    at /tmp/5bf85d1fd7519a4a795c688df1c47749.browserify:6680 <- node_modules/chai/lib/chai/utils/addMethod.js:41:0
	    at /tmp/5bf85d1fd7519a4a795c688df1c47749.browserify:48770 <- tests/filter.js:149:0
```

noahmanger:
Ah yep, I need to update the tests and add new ones. I wanted to hold off to make sure the overall approach made sense, but it sounds like it does, so I'll add them.

noahmanger:
Ok! I fixed the tests. It was a small thing after all. I still want to add new tests, but let's do that after we've finalized the interaction. This should be good to go.

js/filter-control.js
```diff
@@ -4,19 +4,32 @@ var $ = require('jquery');
...
-  this.controlledFilter = $('#' + this.$element.data('controls'));
   this.$element.on('change', this.handleChange.bind(this));
 }
 
+FilterControl.prototype.getFormType = function() {
+  if (this.$element.find('input[type=checkbox]').length > 0) {
+    return 'checkbox';
+  } else if (this.$element.find('select').length > 0) {
+    return 'select';
+  }
```
ccostino: A form within a filter can only ever have checkboxes or a select menu, correct?  If there are forms with both of these elements, is it safe to assume one over the other (in this case, checkbox over select)?

js/filters.js
```diff
@@ -152,6 +214,49 @@ DateFilter.prototype.setValue = function(value) {
...
+    '<div class="message message--error message--small">' +
+      'You entered a date that\'s outside the selected transaction period. ' +
+      'Please enter a receipt date from ' +
+      '<strong>' + this.minYear + '-' + this.maxYear + '</strong>' +
+    '</div>';
+    this.$maxDate.after(warning);
+    this.showingWarning = true;
+  }
+
+  this.$body.trigger('filters:validation', [
```
ccostino: I wonder if this makes more sense to move up into the `validate` method itself.  The reason being is that this is an action to be taken as a result of the validation checks and not necessarily a part of the alert/warning display.  By having it in here, it's coupled with the warning display logic and this may need to change in the future, or we may want the `isValid` state change triggered by something else entirely, independent of showing/hiding a warning.

Does this make sense?  What are your thoughts?

js/filters.js
```diff
@@ -152,6 +214,49 @@ DateFilter.prototype.setValue = function(value) {
...
+  ]);
+};
+
+DateFilter.prototype.hideWarning = function() {
+  if (this.showingWarning) {
+    this.$body.find('.message').remove();
+    this.showingWarning = false;
+  }
+
+  this.$body.trigger('filters:validation', [
```
ccostino: Same note as above in the `showWarning` method.

ccostino:
Thanks, @noahmanger!  I just took a closer look at everything and only had a couple of remarks, let me know if you want to discuss more or need additional information.  Overall the approach looks great to me, especially with the state change and event handler for validation. :-)

js/filter-control.js
noahmanger: Yep. It might be good to abstract this more down the road, but that's all we have at the moment.

js/filters.js
noahmanger: Ah yeah. So just moving line 240 to the `validate` method? That makes sense.

ccostino:
Awesome, thanks @noahmanger!  This is ready for a version bump once merged.

PR #386, opened 2016-06-04
Breakpoint checks should account for scrollbar

### LINKED ISSUE #385: SiteNav::switchMenu does not account for scrollbars
At around ~864px, I see the mobile nav unstyled. Once over the large breakpoint (860px) it should be removed. The reason is the body width calculation doesn't account for scrollbars.
screenshot from 2016-06-02 18-13-33

Here's some background on the issue: https://github.com/jquery/jquery/issues/1729. As of jquery 2.2, you can use `$(window).outerWidth()` to compute this properly.

adborden:
Fixes #385 

Shows up as a revert because I accidentally committed the fix directly to master, then reverted it. This reverts the revert, re-instituting the fix. This actually fixes three things:

- Use `window.innerWidth` to include the scrollbar widths, which will be the same value used for media queries.
- Avoid appending multiple `.js-mobile-nav` elements on every `resize` event.
- Off-by-one breakpoint calculations. 860px should be considered large, because we're using `min-width` in our media queries.

### DIFF
--- a/js/filter-panel.js
+++ b/js/filter-panel.js
@@ -36,7 +36,7 @@ function FilterPanel(options) {
 }
 
 FilterPanel.prototype.setInitialDisplay = function() {
-  if ($(document).width() > helpers.BREAKPOINTS.LARGE) {
+  if (window.innerWidth >= helpers.BREAKPOINTS.LARGE) {
     this.show();
   } else if (!this.isOpen) {
     this.hide();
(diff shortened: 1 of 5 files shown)


noahmanger:
This looks good, except we need to update the tests. To simulate smaller screen sizes, we would set the width of `body

noahmanger:
This looks good, except we need to update the tests. To simulate smaller screen sizes, we would set the width of `body`. Is there a way to force the width of `window` in the tests?

adborden:
Hmm. There is not a good answer to this. In the past, I've usually done responsive testing with webdriver tests which ​*do*​ provide an API to adjust the window size during tests. Karma does not provide this https://github.com/karma-runner/karma/issues/438

To address the tests, I've added a helper around with window width so that we can stub it in tests. It's not ideal, but works. I'd also like to revisit this and maybe replace this with a `helpers.breakpoints.isLarge()` etc.

noahmanger:
Rad. This looks good. Thanks for figuring it out. +1 to moving to `helpers.breakpoints.isLarge()` eventually.

Issue #389, opened 2016-06-07
To avoid the menu from reaching the bottom of the window, restyle spacing & element placement of mega menu


jenniferthibault:
The last addition of the Advanced data button on the mega menu in https://github.com/18F/fec-style/pull/387 eliminated all the vertical space we gained back by designing the smaller header.

We've heard from both FEC stakeholders and at least one user in testing that when the mega menu hits the bottom of the window, it's unclear whether they can scroll within it, or below it, and is adding cognitive load. (There may have been other users, but I only directly observed one). Beyond outside input, our team also recognizes it's not an ideal design experience.

Here's what it currently looks like on a 13in screen:

![tmjnvjuvyo](https://cloud.githubusercontent.com/assets/11636908/15860526/763298b8-2c97-11e6-932f-7e7bd8ee49c5.gif)

I'd like to move this into this sprint, and will assign it to myself, but will also need front end help.

cc @noahmanger @onezerojeremy

jenniferthibault:
Two options for how we could position the `Full advanced data` link to use the space better

**Remove the button, and add "All" to the Advanced Data heading to match others**
- This is my preference, because I think it simplifies it the most, in an already competitive space, and uses the existing language pattern of the child elements. 

<img width="1278" alt="screen shot 2016-06-14 at 12 10 43 pm" src="https://cloud.githubusercontent.com/assets/11636908/16050791/f3a9f94c-322a-11e6-9d73-cb87b4a55e39.png">


**Make a 4th column, and use the `alt` button style to bump it back from the hierarchy a notch. I think it was too prominent among the rest before.**

![screen shot 2016-06-14 at 12 21 43 pm](https://cloud.githubusercontent.com/assets/11636908/16050770/e36a11d4-322a-11e6-88e9-372ef1e5e956.png)

I was tinkering with this in the browser, so the first option is implemented in this PR https://github.com/18F/fec-style/pull/395 but that doesn't mean we have to carry it forward if there are strong reasons for the button approach.

cc @noahmanger @onezerojeremy

noahmanger:
I agree with just adding the "all" to the title. Seems like the simplest solution.

jenniferthibault:
👍 — Moving #395 out of WIP and into review.

PR #395, opened 2016-06-14
Mega menu sizing

jenniferthibault:
## Summary

- Reduces the height of the mega menu so that it avoids hitting the bottom window frame as much as possible
- Resolves https://github.com/18F/fec-style/issues/389

## Screenshots

screen shot 2016-06-14 at 12 10 43 pm

### Review
@noahmanger 

### Tasks to completion

- [x] Agree on approach for integrating link to Adv data page from https://github.com/18F/fec-style/issues/389

### DIFF
--- a/js/templates/nav-data.hbs
+++ b/js/templates/nav-data.hbs
@@ -42,7 +42,7 @@
       </div>
     </div>
     <div class="mega-heading">
-      <h3 class="mega-heading__title icon-heading--table"><a href="{{webAppUrl}}/advanced">Advanced data</a></h3>
+      <h3 class="mega-heading__title icon-heading--table"><a href="{{webAppUrl}}/advanced">Advanced data - all</a></h3>
     </div>
     <div class="mega__group">
       <div class="mega__column">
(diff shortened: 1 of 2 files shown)


js/templates/nav-data.hbs
```diff
@@ -42,7 +42,7 @@
       </div>
     </div>
     <div class="mega-heading">
-      <h3 class="mega-heading__title icon-heading--table"><a href="{{webAppUrl}}/advanced">Advanced data</a></h3>
+      <h3 class="mega-heading__title icon-heading--table"><a href="{{webAppUrl}}/advanced">Advanced data - all</a></h3>
     </div>
```
jenniferthibault: @noahmanger was this the correct file to edit? The changes weren't showing up locally, and I'm not sure familiar with how .hbs works, but it's the only place I could find the text.

noahmanger: Yep this is right. Not sure why it wasn't working. I'll review, but we should be good.

noahmanger:
Works for me! Failing tests aren't actually related to this PR (and are addressed in https://github.com/18F/fec-style/pull/397) so I'm merging.

PR #397, opened 2016-06-16
Fixing filter panel test

noahmanger:
## Summary
- Fixes failing tests caused by `this.originalWidth` returning a value of `384` instead of a desktop-sized width. (Not sure why it was doing this).
- While investigating this, I realized we weren't setting `aria-hidden` correctly on the filter panel, so I fixed that as well.

cc @adborden


adborden:
The `originalWidth` issue is likely because another test is changing the body width but not resetting it in the test cleanup.

PR #398, opened 2016-06-16
fixing rem calc bug

anthonygarvan:
I am fixing this in one place (where it was causing me an issue) but it may be a widespread problem. Anyplace the a `calc(X - Yrem)` is used in order to offset by an element of width `u(Yrem)`, the offset will be broken when you adapt the mode (eg,. fec-eregs sets $px-only to `true`, which surfaced this bug).

### DIFF
--- a/scss/components/_search-bar.scss
+++ b/scss/components/_search-bar.scss
@@ -22,6 +22,8 @@
 //
 // Styleguide components.search-bar
 
+$search-button-width: u(5.6rem);
+
 .search__select {
   border-radius: 4px 0 0 4px;
   margin-right: -2px;
@@ -42,7 +44,6 @@
   .combo__input {
     border-radius: 4px 0 0 4px;
     float: left;
-    $search-button-width: u(5.6rem);
     width: calc(100% - #{$search-button-width});
 
     .combo__input {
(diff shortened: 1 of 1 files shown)


scss/components/_search-bar.scss
```diff
@@ -42,7 +42,8 @@
   .combo__input {
     border-radius: 4px 0 0 4px;
     float: left;
-    width: calc(100% - 5.6rem);
+    $search-button-width: u(5.6rem);
```
noahmanger: I like assigning this to a variable (in addition to solving the bug). What about also using it when defining the width of the button? Also, can we move the variable assignment to the top of the list?

adborden:
Nice, Tony, there are a handful of places where we use `calc` but I wasn't able to solve using the `u()` function on it's own. Good thinking! I wonder if we can just create a `calc()` function/mixin to do this automatically for us.

anthonygarvan:
I think as @noahmanger points out, it's probably good to name these variables anyway for clarity.

scss/components/_search-bar.scss
```diff
@@ -76,7 +78,7 @@
     }
 
     .combo__button {
-      width: u(5.6rem);
+      width: #{$search-button-width};
```
adborden: Shouldn't need the string interpolation here, you can just use the variable directly:
```
width: $search-button-width;
```

PR #402, opened 2016-06-21
Tooltip adjustments

noahmanger:
## Summary
Adjusts the way that tooltips should be marked up in order for them to be positioned more consistently and for the point to always be under the question mark trigger (if present). Specifically, the `.tooltip__container` class should now wrap the `.tooltip_trigger` and `.tooltip` only, not the text associated with the trigger as it previously was.

Along these lines, I replaced the CSS-generated arrow with an SVG, as in order to get the border effect on the CSS arrow we essentially had to overlap a blue and white triangle. For whatever reason, they would often be misaligned in some places but not others, and the simplest solution was to just make a static asset.

This also contains some minor spacing adjustments to the calendar list view.


noahmanger:
cc @adborden on this. Are you using tooltips at all on legal stuff? If so, this introduces a semi-breaking change.

adborden:
No, not using tooltips explicitly, although they might appear in some of the FEC components we're using.. (but I can't think of any).

PR #404, opened 2016-06-21
Using display:table-cell for height matching

noahmanger:
## Summary
This removes the JS that we were using to set the filter panel height in favor of using some old-school CSS `display: table-cell` magic. I can't remember if we had a good reason for not doing this earlier, but it seems to work just fine.

What do you think @adborden ?

## Screenshots
image

Resolves https://github.com/18F/fec-cms/issues/319


adborden:
Awesome!

cc @anthonygarvan

PR #406, opened 2016-06-22
Adds operating expenditures

emileighoutlaw:
Adds operating expenditures, which we'll want to have available when we launch the spending overview pages. This definition is approved by Amy K. and pulled directly from the campaign guide. 

_cc @jenniferthibault _


jenniferthibault:
Nice!

PR #409, opened 2016-06-22
Re-sizes and positions the card icons

noahmanger:
## Summary
Re-sizes and re-positions the icons in the secondary cards on reporting and registration pages:
image

Also renames the `.button--cal` class to `.button--grid` and uses a different icon for the grid-view toggle on the calendar page to avoid the conflict with the larger calendar icon used in cards.

cc @jenniferthibault


jenniferthibault:
👀

PR #412, opened 2016-06-25
Datatable styles

noahmanger:
## Summary
This introduces a few data table style changes, namely:
- Tiger-striped table cells
- Right border on data table cells
- Doesn't stick the sticky bar on small screens
- A new `.column--number` class that right-aligns columns with numbers (dates and money)
- New size classes for data-table columns to control the sizing in CSS rather than in the JS that defines each table

I'll post screenshots after things have settled. Still a WIP.

### DIFF
--- a/js/sticky-bar.js
+++ b/js/sticky-bar.js
@@ -14,16 +14,16 @@ function StickyBar(selector) {
 }
 
 StickyBar.prototype.toggle = function() {
-  if (helpers.isLargeScreen()) {
-    var scrollTop = this.$body.scrollTop();
-    if (scrollTop >= this.offset + this.triggerOffset) {
-      var height = this.$bar.outerHeight();
-      this.$bar.addClass('is-stuck');
-      this.$body.css('padding-top', height);
-    } else if (scrollTop < this.offset) {
-      this.$bar.removeClass('is-stuck');
-      this.$body.css('padding-top', this.defaultBodyPadding);
-    }
+  if (!helpers.isLargeScreen()) { return; }
+
+  var scrollTop = this.$body.scrollTop();
+  if (scrollTop >= this.offset + this.triggerOffset) {
+    var height = this.$bar.outerHeight();
+    this.$bar.addClass('is-stuck');
+    this.$body.css('padding-top', height);
+  } else if (scrollTop < this.offset) {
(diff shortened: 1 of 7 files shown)


js/sticky-bar.js
```diff
@@ -13,14 +14,16 @@ function StickyBar(selector) {
...
 StickyBar.prototype.toggle = function() {
-  var scrollTop = this.$body.scrollTop();
-  if (scrollTop >= this.offset + this.triggerOffset) {
-    var height = this.$bar.outerHeight();
-    this.$bar.addClass('is-stuck');
-    this.$body.css('padding-top', height);
-  } else if (scrollTop < this.offset) {
-    this.$bar.removeClass('is-stuck');
-    this.$body.css('padding-top', this.defaultBodyPadding);
+  if (helpers.isLargeScreen()) {
```
adborden: nit picky, but I prefer a "fail fast" style rather than the entire body of the function in an `if`... e.g.
```javascript
if (!helpers.isLargeScreen()) {
  return;
}

// function code
```

scss/modules/_datatables.scss
```diff
@@ -253,3 +278,42 @@
...
+    .column--xl,
+    .column--large {
+      width: 50%;
+    }
+
+    .column--med {
+      width: 25%;
+    }
+  }
+}
```
adborden: Hmm, good call on the column sizing. I had added the `.cell--5` classes, but to have a standard set seems better.

adborden:
:+1: @jenniferthibault did you want to take a look?

js/sticky-bar.js
noahmanger: Ah! I like it. Fixed.

scss/modules/_datatables.scss
noahmanger: Yeah I struggled with if it made sense to use those classes or these. I ultimately decided the more generic ones made sense because they'd be different sizes at different screen sizes. I'm not opposed to keeping the others in here if they're useful, but it might be good to settle on an approach. Do you have strong feelings one way or the other?

noahmanger:
Jen already gave it the 👍  so it's good to merge whenever you're happy with it.

Issue #391, opened 2016-06-09
Mobile menu assumes positioned container is full height


adborden:
Mobile menu height comes from it's absolute positioning with `bottom: 0`. Since it's positioned container is the `body` element, the menu is full height. On eRegs, the positioned container can not easily be made to be the `body` element, so it's positioned container (likely the fixed header) doesn't provide enough space for the mobile nav. The goal is to size the menu without any eRegs specific overrides.

**Mobile menu on eRegs after overriding CSS to make the menu the height of its contents.**
![screenshot from 2016-06-09 14-40-09](https://cloud.githubusercontent.com/assets/509703/15947419/39417220-2e50-11e6-90e1-a3420c465658.png)

**Mobile menu on betaFEC**
![screenshot from 2016-06-09 14-39-45](https://cloud.githubusercontent.com/assets/509703/15947398/231c48da-2e50-11e6-8253-2b1331e7500b.png)

Thoughts on possible solutions:
- Size the menu full height with javascript
- Specify the height by its content
- Specify a fixed height

adborden:
cc @noahmanger

Issue #390, opened 2016-06-09
Mobile menu sometimes includes utility items, sometimes not


adborden:
If you open the FEC site on mobile, the Nav has 6 items, including the utility items About, Contact Us, and Calendar. When you open the FEC site on desktop, then shrink it down to mobile breakpoints, the nav menu does not include these times.

Opening on mobile:
![screenshot from 2016-06-08 17-34-35](https://cloud.githubusercontent.com/assets/509703/15915636/e489583e-2d9f-11e6-8698-504244761331.png)

Opening on desktop, resizing to mobile:
![screenshot from 2016-06-08 17-35-19](https://cloud.githubusercontent.com/assets/509703/15915644/ef1475cc-2d9f-11e6-9f02-20cc49ed4dc8.png)

PR #414, opened 2016-06-30
Mobile menu fix

noahmanger:
## Summary

This fixes a few problems under the hood with the mobile menu:
- Sets position relative to the header, rather than the body, which is better for eregs
- Fixes a bug where utility nav items and the menu title weren't showing up if you started at a big display and then shrank it

In order to fix that second one, I forked the mega menu library and [edited it slightly](https://github.com/noahmanger/Accessible-Mega-Menu/commit/969e51be809bb6a4a481befc0c31ca8089c2c945) to allow for a selector to be passed through to designate which nav items would actually contain mega menus. This solves the problem where the mega menu expected every `li` to contain an item, so was appending various ARIA attributes that they shouldn't have (like `aria-hidden`), which was causing them to not show up. This has the added benefit of not including the menu library in this repo.

I decided to go with a fork rather than suggesting an upstream fix because they don't seem to be maintaining that repo anymore.

Resolves https://github.com/18F/fec-style/issues/391
Resolves https://github.com/18F/fec-style/issues/390

cc @adborden


tests/site-nav.js
```diff
@@ -46,11 +46,11 @@ describe('SiteNav', function() {
 
   describe('constructor()', function() {
     it('should set body to jqueryized selector', function() {
-      expect(this.siteNav.$body).to.be.ok;
-      expect(this.siteNav.$body.length).to.be.ok;
```
adborden: These tests are still valid, right? (we still have a `$body` property).

js/site-nav.js
```diff
@@ -113,19 +117,21 @@ SiteNav.prototype.toggleMenu = function() {
...
+  this.$body.css({'overflow': 'hidden'});
+  this.$element.addClass('is-open');
   this.$toggle.addClass('active');
   this.$menu.attr('aria-hidden', false);
   this.isOpen = true;
 };
 
 SiteNav.prototype.hideMenu = function() {
-  this.$body.removeClass('is-open');
+  this.$body.css({'overflow': 'auto'});
```
adborden: Would be good to add tests that check the `overflow` is toggled.

adborden:
:+1: What's the plan for the mega menu? Do you think you/18F should take ownership of it? Would be nice to get it into npm proper. Maybe you should open an issue in their repo asking if they are going to continue support for it?

tests/site-nav.js
noahmanger: Yeah, I just renamed it to `$element`. The test stayed the same.

noahmanger:
Good question on the mega menu. I'm not sure what's best. I'll ask in Slack.

js/site-nav.js
noahmanger: 👍

tests/site-nav.js
adborden: Oh, I mean that these lines could stay as is, having the same test for `siteNav.$body` _and_ for `siteNav.$element`.

PR #419, opened 2016-07-06
Fixing analytics

noahmanger:
This addresses a problem caused by the fact that we have two Google Analytics trackers on the site (one for the Digital Analytics Program and one for non-DAP GA) where custom events that didn't have a tracker specified would throw an error and our analytics would not register custom events.

So this addresses it by:
- Only adding the GA script if it hasn't already been added by DAP
- Adding a name (`nonDAP`) to the non-DAP tracker
- Calling `set` and `send` with the new namespaced tracker

This goes with a companion PR on the web app https://github.com/18F/openFEC-web-app/pull/1339 .

I'd like @gbinal 's eyes on this if you have time and maybe @jmcarp, who originally set up the analytics code on the project. 

It definitely solved the problem, but I want to make sure I'm not inadvertently breaking something else. Using the GA Debugger for Chrome it looks like the DAP events are still being sent and everything, but I just want to be safe.

### DIFF
--- a/js/analytics.js
+++ b/js/analytics.js
@@ -5,15 +5,26 @@
 var URI = require('urijs');
 var _ = require('underscore');
 
-function init() {
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+function trackerExists() {
+  if (typeof ga !== 'undefined') {
+    return true;
+  }
+}
 
-  ga('set', 'forceSSL', true);
-  ga('set', 'anonymizeIp', true);
-  ga('create', 'UA-48605964-22', 'auto');
+/* Initialize a non Digital Analytics Program GA tracker
+ * This tracker's name is "nonDAP", so all commands will need to be prefixed
+*/
+function init() {
+  if (!trackerExists) {
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+    (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+    m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+  }
+
+  ga('create', 'UA-48605964-22', 'auto', 'notDAP');
+  ga('notDAP.set', 'forceSSL', true);
+  ga('notDAP.set', 'anonymizeIp', true);
 }
 
 function pageView() {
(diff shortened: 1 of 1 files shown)


gbinal:
Unfortunately, on this I'm not in a position to offer a useful 👍 .  Sorry not to be more help.  It might make sense to holler in #g-analytics if you need another pair of 👀.

noahmanger:
No worries. Will do.

js/analytics.js
```diff
@@ -5,15 +5,26 @@
...
 function init() {
-  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
-  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
-  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
-
-  ga('set', 'forceSSL', true);
-  ga('set', 'anonymizeIp', true);
-  ga('create', 'UA-48605964-22', 'auto');
+  if (!trackerExists) {
```
adborden: Eep, this should be `trackerExists()`

adborden:
This seems a bit funky. I would expect #g-analytics to have some advice on how to handle having DAP and plain old GA together. There's an implicit dependency on the order of loading between DAP and GA, since when this code runs, it's only then that we detect if `ga` exists or not, and it's unclear what DAP does when it's first vs second, unless you're familiar with the DAP code.

js/analytics.js
noahmanger: Ack! Good catch.

adborden:
As a point of reference, `regulations-site` allows you to set both GA and DAP in which case it loads the GA code and then the DAP code.
https://github.com/adborden/regulations-site/blob/ebf6c318d2ff6a6d82e999ab518b4fd0be5a8275/regulations/templates/regulations/base.html#L57
https://github.com/adborden/regulations-site/blob/ebf6c318d2ff6a6d82e999ab518b4fd0be5a8275/regulations/templates/regulations/google_analytics.html#L17
https://github.com/adborden/regulations-site/blob/ebf6c318d2ff6a6d82e999ab518b4fd0be5a8275/regulations/templates/regulations/dap_analytics.html#L6

js/analytics.js
```diff
@@ -5,15 +5,26 @@
 var URI = require('urijs');
 var _ = require('underscore');
 
+function trackerExists() {
+  if (typeof ga !== 'undefined') {
+    return true;
+  }
+}
```
adborden: For this function, which I didn't explain very well, the idea was that `ga` is not used by any other code, it would only be used within `analytics.js`. `analytics` could expose a `track` function, would check if `ga` exists and then call `ga`. If it's not available, then `track` is a no-op. This means that external code always calls `analytics.track` and never has to check whether `ga` exists or not, it lets `track` decide what to do.

adborden: ... and as long as the consumer calls `init`, `ga` will always exist, so I'm not totally sure if there was a strong reason to check `ga` existing in the first place. Were you seeing something different at some point?

noahmanger: Ah that makes sense. Thanks for clarifying. I didn't set up the check for `ga` originally, but as far as I can tell it was there because external code was calling methods so that code had to check first.

adborden:
I'm good to merge this as-is. The fact that `trackerExists` isn't getting called is probably a good thing, because it's consistently creating `ga` which seems to be working in testing. We'll come back next release to clean this up :+1:

PR #423, opened 2016-07-08
Adding a slight delay to the nav opening

noahmanger:
## Summary

I discovered a PR to the accessible mega menu library that someone made but which never got merged in that adds the ability to define a delay to the opening of the mega menu. I tried it out and have it queued up as a PR into my fork: https://github.com/noahmanger/Accessible-Mega-Menu/pull/1

This PR just makes use of that param. I know we're still tinkering with the concept of the mega menu, but this seems like a nice UI tweak to make in the mean time. 

What do you think @jenniferthibault @onezerojeremy (for UX) and @adborden ?

### DIFF
--- a/js/site-nav.js
+++ b/js/site-nav.js
@@ -79,6 +79,7 @@ SiteNav.prototype.initMegaMenu = function() {
     hoverClass: 'is-hover',
     focusClass: 'is-focus',
     openClass: 'is-open',
+    openDelay: 350,
     selectors: {
       topNavItems: '[data-submenu]'
     }
--- a/js/site-nav.js
+++ b/js/site-nav.js
@@ -79,7 +79,7 @@ SiteNav.prototype.initMegaMenu = function() {
     hoverClass: 'is-hover',
     focusClass: 'is-focus',
     openClass: 'is-open',
-    openDelay: 350,
+    openDelay: 500,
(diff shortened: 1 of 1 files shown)


adborden:
Still interested in getting this change in?

noahmanger:
I am, yeah! What do you think? @onezerojeremy ?

noahmanger:
Pinging @onezerojeremy @nickykrause @jenniferthibault : seeing a number of comments in usability testing and elsewhere (and my own personal annoyance) with the mega menu opening suddenly reminds me that I've had this in the queue for a while. It adds a half-second delay to the opening of the menu so you're less likely to inadvertently open it. 

It's a little hard to tell in the video, but take a look: http://recordit.co/gxnjeeuOBA

I know we still want to revisit the navigation concepts as a whole, but this seems like a good fix to get in in the mean time.

jenniferthibault:
I can't really tell the difference from the video, but +100 this drives me nuts.

onezerojeremy:
Good catch! I think 1/2 second sounds about right, and it looks right too.

noahmanger:
@adborden this also requires merging in the change to the actual menu library. Can you take a look? https://github.com/noahmanger/Accessible-Mega-Menu/pull/1

PR #424, opened 2016-07-08
Filter accessibility

noahmanger:
## Summary
Makes a couple changes to fix some accessibility issues with our filters:

Disables "hints" on the typeahead filters. We weren't using these in any of the other typeahead search fields, and they were posing a minor accessibility issue as it entailed adding an input on top of the input where you were entering text, but this input was both a keyboard focus trap and also lacked a proper label. So instead of this:
image

We'll have this:
image

It also accomodates changes coming on the web app and cms where `<form>` elements that wrapped the filters are replaced with `<div>`s, since there's no submit button. This required a change to the `filterSet` component to have `serializeArray()` called on the individual elements themselves, rather than the `form`, since it cannot be called on a `div`.

Resolves https://github.com/18F/openFEC-web-app/issues/1310
Resolves https://github.com/18F/openFEC-web-app/issues/958

### DIFF
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -6,8 +6,6 @@ var URI = require('urijs');
 
 var Filter = require('./filters').Filter;
 
-var KEYCODE_ENTER = 13;
-
 function FilterSet(elm) {
   this.$body = $(elm);
   $(document.body).on('tag:removed', this.handleTagRemove.bind(this));
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -39,7 +39,7 @@ FilterSet.prototype.activate = function() {
 };
 
 FilterSet.prototype.serialize = function() {
-  return _.reduce(this.$body.serializeArray(), function(memo, val) {
+  return _.reduce(this.$body.find('input,select').serializeArray(), function(memo, val) {
     if (val.value && val.name.slice(0, 1) !== '_') {
       if (memo[val.name]) {
         memo[val.name].push(val.value);
--- a/scss/components/_search-bar.scss
+++ b/scss/components/_search-bar.scss
@@ -205,6 +205,10 @@ $search-button-width: u(5.6rem);
   border-bottom: 1px solid $base;
 }
 
+.tt-highlight {
+  background-color: rgba($aqua, .3);
+}
+
 // Specific styles for typeahead filters
 // Just testing things out. Refactor to use non-js class if we use it
 .js-typeahead-filter {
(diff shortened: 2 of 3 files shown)


jenniferthibault:
@noahmanger If we are making this change, can I suggest we add an improvement (in this, or another PR) that introduces text highlighting between entered text and the matches below?

This is a quick hack with an overlay, but gets the idea across:

screen shot 2016-07-08 at 7 36 10 pm

noahmanger:
Well that was easy! What value do you want the highlight to be? Here it is with `$aqua` at 30% opacity.

image

jenniferthibault:
I dig it! Nice pairing with the bold. I chose the aqua to pair with the current selection highlight bar on the left, and the opacity here works well for me. Visible but not stronger than the text.👌 Let's roll w/ it. 🚀

scss/components/_search-bar.scss
```diff
@@ -205,6 +205,10 @@ $search-button-width: u(5.6rem);
   border-bottom: 1px solid $base;
 }
 
+.tt-highlight {
+  background-color: rgba($aqua, .3);
```
xtine: 0.3?

js/filter-set.js
```diff
@@ -6,8 +6,6 @@ var URI = require('urijs');
 
 var Filter = require('./filters').Filter;
 
-var KEYCODE_ENTER = 13;
```
xtine: What was the purpose of this?

noahmanger: I can't recall, but I think it's because we used to have a handler for keydown that was using that keycode. The var isn't referenced anymore in the file though, so I deleted it.

scss/components/_search-bar.scss
noahmanger: Yeah, we just wanted a subtle highlight of `$aqua` at 30%.

xtine: To clarify I'm wondering if there should be a leading zero on a decimal point or if a shortened style is preferred.

noahmanger: Oh! Sorry. We haven't been using leading zeros elsewhere. Is there a reason to?

xtine: Not necessarily! Just wanted to make sure convention was being followed, sorry my OCD. :)

xtine:
👍

scss/components/_search-bar.scss
noahmanger: No apologies necessary! Good eye.

PR #425, opened 2016-07-11
Adding ethnio signup to feedback form

noahmanger:
## Summary
Since @anthonygarvan [figured out](https://github.com/18F/fec-cms/pull/382/) how to force Ethnio to open, we can now add a sign-up link and button to the feedback form. 

To do that, I added a new `ethnio-helpers.js` file with a couple functions to open the screener and to load the script if it hadn't already been loaded.

A couple notes though:
- Tony's original implementation just opened the first popup box (the one that opens in the corner) and then you need to click through that to get to the screener form; this one opens up the screener directly (but the corner box has to open first). The advantage to the first approach is that you get the intro text from the initial popup. The advantage to mine is that it opens the screener faster.
- I'm running into an issue where after I open the screener, if I try to X it out it thinks it's a successful submission and then I can't close it. Is anyone else having this problem?

Resolves https://github.com/18F/openFEC-web-app/issues/1143

## Screenshots
from main screen

from success

cc @anthonygarvan because I'm interested in what you think of the implementation and @jenniferthibault and @onezerojeremy for the UX questions.


noahmanger:
@jenniferthibault do you have thoughts on this interaction?

jenniferthibault:
I'm a little confused about what's supposed to happen, and what the ux problems to solve are. Let's go synchronous on this one? Calendaring now.

noahmanger:
After talking about this with @jenniferthibault and @onezerojeremy , we decided the best approach is to just have the links go to a separate web page, rather than trigger the pop up. I've updated the PR to do just that, and this is now ready for review.

xtine:
👍

PR #434, opened 2016-07-18
Filter tab bug

noahmanger:
## Summary
- Only show the "Clear all filters" button if the tags are removable. That way, in cases where certain filters are applied by default and not removable, it doesn't give the option to remove them.
- Adds a non-removable filter tag for "transaction period" that follows this logic

## Screenshots
image

Resolves https://github.com/18F/openFEC-web-app/issues/1311

cc @xtine


xtine:
Ha! I found this bug while playing around with different filters and was going to file an issue. 😉
Nice work! 👍

PR #435, opened 2016-07-19
Filter loading states

noahmanger:
## Summary
I had some time so I decided to get a head start in order to figure out where the different pieces are going to live. So this:

- Adds `is-loading`, `is-successful` and `is-unsuccessful` classes for checkboxes and radio buttons
- Adds logic to `filters.js` to set the loading state on checkboxes and radios 
- Only adds the class if the filters have been loaded already once for the page load

@jenniferthibault I realized we didn't discuss error styles for these, so I took a best guess. The scenario where these would be used is if you apply a filter but an error happens in loading the table. This is unlikely, but probably good to specify.

## Screenshots
image
image
image

cc @xtine


jenniferthibault:
I added one commit to mark the USWDS colors we are using, and one question about the variable names.

On the error state, that makes sense, but we'd want to make sure to have an error message that accompanies it, in the same way we have a confirmation message for the others.

scss/_variables.scss
```diff
@@ -25,12 +25,15 @@ $navy: #164f85;
...
 $gray-light: #e8e8e8;
-$gray-medium: #d6d7d9; // U.S. Web Design Standards color gray-mediumer //
+$gray-medium: #d6d7d9; // U.S. Web Design Standards color gray-lighter //
 $gray: #aeb0b5; // U.S. Web Design Standards color gray-medium //
 $gray-dark: #5b616b; // U.S. Web Design Standards color gray //
 
 
 $disabled: #737373;
+$green-light: #4AA564; // U.S. Web Design Standards color green-light //
+$green-dark: #287538;
```
jenniferthibault: @noahmanger if the red below is `$error`, should these greens be named something like `$success` or `$confirm` ?

scss/components/_dropdowns.scss
```diff
@@ -33,6 +33,22 @@
     top: u(.5rem);
   }
 
+  &.is-loading {
+    &::after {
+      background-image: url('../img/loading-ellipsis-gray.gif');
+      background-size: 18px;
```
noahmanger: Let's use rem here.

noahmanger: Or 75% like in `.button--loading`

scss/components/_dropdowns.scss
```diff
@@ -33,6 +33,21 @@
...
+      padding-left: u(1.5rem);
+      right: u(1rem);
+    }
+  }
```
noahmanger: ☝️

scss/components/_dropdowns.scss
```diff
@@ -62,7 +78,7 @@
   overflow: hidden;
   text-align: left;
   top: u(3.4rem);
-  z-index: 999;
+  z-index: 10;
```
noahmanger: My bad, since I originally wrote the 999, but let's use a z-index variable here.

scss/components/_form-styles.scss
```diff
@@ -219,3 +219,61 @@
 .date-range-input {
   padding-left: u(2.5rem);
 }
+
+// Loading states
+label {
```
noahmanger: Should we be specifying this further so as not to target all `labels`?

scss/components/_form-styles.scss
```diff
@@ -219,3 +219,61 @@
...
+
+  &::after {
+    background: transparent;
+    content: '';
+    display: inline-block;
+    width: 2rem;
```
noahmanger: I think this is my fault, but these should all be `u(...)`. This is necessary because the styles are shared by the FEC instance of eRegs.

scss/elements/_forms.scss
```diff
@@ -177,6 +177,12 @@ input[type="file"] {
   }
 }
 
+// set smaller width for checkbox label
```
noahmanger: I never explained this, but all of the `/elements/` partials should be unclassed, html element selectors only. So let's move this style to the `_dropdowns.scss` partial.

scss/modules/_filters.scss
```diff
@@ -145,6 +155,39 @@
     height: auto;
   }
 
+  .combo--filter--mini {
```
noahmanger: Not sure: does this make more sense here or in `_form-styles.scss` with the other `.combo` styles? I _think_ there, but not sure.

js/filters.js
```diff
@@ -33,15 +33,24 @@ function Filter(elm) {
   this.$body = $(elm);
   this.$input = this.$body.find('input:not([name^="_"])');
   this.$remove = this.$body.find('.button--remove');
+  this.$inputFilter = this.$body.find('.input--removable input');
```
noahmanger: Maybe a topic for a separate PR, but I wonder if we should break out the text input filters from the checkbox filters into their own classes, so that we don't have as much logic here?

noahmanger:
This works well except for one weird thing I noticed: when hitting enter on a loading button, the background image is huge:

weird

The rest of my comments are mostly picky style things or have to do with the best place for code to live. Feel free to push back on any.

scss/modules/_filters.scss
xtine: i think it's a case that it would make sense in either folder. however it seems `filters.scss` contains most of the interaction classes and styles now

xtine:
I addressed most of the issues noted, @noahmanger let me know if I've missed anything.

I'm guessing the large loading ellipsis image seems to be an image preloading issue. We might need to cache it beforehand because because after an initial load I didn't get the huge background loading image.

PR #438, opened 2016-07-27
Full-height filter panel on mobile

xtine:
## Summary
- Changed filter panel to slide from the left over data table in mobile.

## Screenshots
t3sn2ydqz1

18F/openFEC-web-app#1326

/cc @noahmanger @jenniferthibault

### DIFF
--- a/js/filter-panel.js
+++ b/js/filter-panel.js
@@ -43,13 +43,17 @@ FilterPanel.prototype.setInitialDisplay = function() {
   }
 };
 
-FilterPanel.prototype.show = function() {
+FilterPanel.prototype.show = function(focus) {
   this.$body.addClass('is-open').attr('aria-hidden', false);
   this.$toggle.attr('aria-hidden', true);
   accessibility.restoreTabindex(this.$form);
   $('body').addClass('is-showing-filters');
   this.isOpen = true;
-  this.$body.find('input').first().focus();
+  // Don't focus on the first filter unless explicitly intended to
+  // Prevents the first filter from being focused on initial page load
+  if (focus) {
+    this.$body.find('input, select, button:not(.js-filter-close)').first().focus();
+  }
 };
 
 FilterPanel.prototype.hide = function() {
@@ -65,7 +69,7 @@ FilterPanel.prototype.toggle = function() {
   if (this.isOpen) {
     this.hide();
   } else {
-    this.show();
+    this.show(true);
   }
 };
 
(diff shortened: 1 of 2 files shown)


noahmanger:
Is this ready for review?

js/filter-panel.js
```diff
@@ -49,7 +49,7 @@ FilterPanel.prototype.show = function() {
   accessibility.restoreTabindex(this.$form);
   $('body').addClass('is-showing-filters');
   this.isOpen = true;
-  this.$close.focus();
+  this.$body.find('input').first().focus();
```
noahmanger: This should probably be more generic to include `select` elements as well, in the event it's the first one.

scss/modules/_filters.scss
```diff
@@ -20,25 +20,34 @@
...
   @include u-bg--neutral();
+  @include transition(left 1s ease-in-out);
+  position: absolute;
   bottom: 0;
-  height: 0;
-  top: 0;
+  height: auto;
+  top: -20px;
+  left: -800px;
+  z-index: 10;
```
noahmanger: We use variables for z-indexes. They're all in `_variables.scss` at the bottom. You'll want to choose one that's greater than the one used for $z-overlay` because that's the loading state overlay, which is currently showing up on top of the filter panel when it's open.

scss/modules/_filters.scss
```diff
@@ -20,25 +20,34 @@
 //
 // Styleguide modules.filters
 
+// by default show filters offscreen
 .filters {
   @include clearfix();
   @include u-bg--neutral();
+  @include transition(left 1s ease-in-out);
```
noahmanger: Let's speed up the transition and go with ease-out to match the transition of the mobile menu:
`.2s ease-out`

scss/modules/_filters.scss
```diff
@@ -20,25 +20,34 @@
...
+  position: absolute;
   bottom: 0;
-  height: 0;
-  top: 0;
+  height: auto;
+  top: -20px;
```
noahmanger: This should be `rem` because the padding of the element that it's compensating for is set in `rem` (and it's the convention we've been using on the project).

noahmanger:
Working super well. Just had a few little comments.

js/filter-panel.js
xtine: Are there `select` elements? Because the dropdowns are currently hidden input checkboxes.

noahmanger: There are with the transaction period filters. These aren't first, so this was more a thought about future-proofing than addressing a current need.

noahmanger: I think just `this.$body.find('input,select').first()` should work, right?

xtine:
Cool, thanks for that extra focus bit @noahmanger.

PR #439, opened 2016-07-28
Show number of applied filters on accordion heading

xtine:
## Summary
On filter accordion headings:
- shows number of applied filters upon load
- when a filter is applied, number increases by 1
- when a filter is removed, number decreases by 1
- when there are no filters, no numbers are shown.

Needs tests.

18F/openFEC-web-app#1327

## Screenshots
mr5zzvvolb

/cc @noahmanger


xtine:
Closing because @noahmanger got this working better in #440.

PR #440, opened 2016-07-28
Adding filter count incrementing to filters.js

noahmanger:
Adds the filter count incrementing and decrementing from https://github.com/18F/fec-style/pull/439


xtine:
👍 Keeping track of initially loaded data is great.

PR #441, opened 2016-07-28
Data landing mobile

noahmanger:
## Summary
- Adds separate styles for the data landing page chart and totals for medium-sized screens, so they're next to each other like they are on desktop
- Adds separate styles for when the tooltip point is it's own element, rather than a psuedo-element, which is useful for positioning it with javascript


xtine:
👍

PR #442, opened 2016-07-28
Adding tests for incrementing filter count

noahmanger:
## Summary
Adds tests for incrementing and decrementing filter count. 

I decided to just do this as a PR so it'd be easier to review.


xtine:
Tests make sense and pass, good stuff. 👍

PR #445, opened 2016-08-01
Calendar tooltips

noahmanger:
## Summary
Styles the calendar category tooltips like so:

image

I totally forgot to file a PR for this earlier. It should be going into this release.

cc @xtine

### DIFF
--- a/scss/components/_tooltips.scss
+++ b/scss/components/_tooltips.scss
@@ -48,6 +48,10 @@
     border-bottom-color: $primary;
     color: $primary;
   }
+
+  li {
+    font-size: u(1.4rem);
+  }
 }
 
 .tooltip--above {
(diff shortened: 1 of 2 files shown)


noahmanger:
Mmm kinda. So it's just the right border on the text element:

image

The way to increase space is by just adding padding to the overall box. Currently it has 2px top and bottom padding. That screenshot there is what it looks like with 5px. Want more?

jenniferthibault:
Iiiiiinteresting. Thanks for explaining how it's built!

I guess I'd prefer to keep it as it was then, since increasing padding to the box makes it look more like a button than a tag. 

My overall observation is that the pipe feels tall relative to the copy, since it's hitting the top and bottom of the copy box (which probably includes margin?), not just the limit of ascender/descenders.

ccostino:
@noahmanger and @jenniferthibault:  It looks like this is good to go as is then, correct?  All looks fine to me, I just wanted to confirm!

noahmanger:
Yep!

xtine:
Cool, just wanted to wait until design was resolved. LGTM. 👍

PR #446, opened 2016-08-02
New filter panel layout

noahmanger:
## Summary
New mobile and desktop styles for the filter panel, [as outlined here](https://github.com/18F/openFEC-web-app/issues/1379). Included here:
- Adds new style for the filter panel toggle button
- Adds new styles for the filter panel, so that now the `.filter__content` element is hidden, not `.filters` (because `.filters` has the toggle button, which is always visible)
- Now the filter toggle button shows the number of applied filters (handled by the `FilterPanel` component)
- The `TagList` component is updated with new language and markup so that the `DataTable` component can add the number of results to the title:
- Adds a new style for `.button--unstyled` which is used by the clear all tags button
- Hides the `.page-header` element on mobile (which contains the breadcrumbs on larger screens)
- Restyles the filter tags
- Updates calendar styles for the new layout

## Screenshots
image
image
image

@jenniferthibault paired with me for visual review, so this is ready for code review. Probably good for both @xtine and @adborden to be aware of the changes. 

Goes with https://github.com/18F/fec-cms/pull/424 and https://github.com/18F/openFEC-web-app/pull/1388


noahmanger:
I updated the original comment with an explanation of what's included. Not sure if people get pings when that's updated, so cc @xtine @adborden and @jenniferthibault .

jenniferthibault:
Took a cruise through and caught a coupl'a things:

- When you load the page, the filter panel is in a between-closed-and-open state. Is it possible to get it to settle on one or the other? Starting closed then sliding out to open would be fine, if that's easier.

 qzijfkdhcx

- On the calendar, when the window gets smaller, the icons, pipe, and button dropdowns crash into each other

 screen shot 2016-08-04 at 10 53 43 am

- When you filter down to zero results, the text says `Viewing about "Showing 0 to 0 of 0 entries" filtered results for` which is a mashup of the old and new structures.
While we clean this up, is it possible to use a totally different text string for when the results have all been filtered away? Something like `Your filters do not produce any results`cc @emileighoutlaw or would we be stuck in the `Viewing about "0" filtered results for:` format? It seems weird to say "about 0"

screen shot 2016-08-04 at 9 49 25 am


- [x] made filter icon 10% larger inside the circle (already done in https://github.com/18F/fec-style/pull/446/commits/440ad80d38ce27af13938a5b5361acabc29f49ae)

screen shot 2016-08-04 at 11 04 58 am

jenniferthibault:
ooh one more. I'm not seeing the "Edit filters" placed below the page title (as it should be in the screenshot below) but I remember seeing this in your local version @noahmanger , so just wanted to confirm

screen shot 2016-08-04 at 11 35 03 am

noahmanger:
Another to-do:
- [ ] Make sure the filter count subtracts on filter removal

noahmanger:
Fixed all of the above. For the 0 results count, let's keep it simple and just show 0 for now:
image

For the filter button not showing up in the right order, did you refresh the page after adjusting the window?

This is good to go.

adborden:
:+1:

PR #447, opened 2016-08-03
Position icons 1rem from right on certain buttons

noahmanger:
For buttons with the icons on the right, rather than positioning those background images at `90%` from the left, this positions them at `1rem` from the right.

cc @porta-antiporta


xtine:
Just for reference, what icons do these affect? Can you post a screenshot?

noahmanger:
Oops, sorry. It's all the buttons with icons on the right: 
image

PR #452, opened 2016-08-04
Filter loading states for typeahead input

xtine:
## Filter loading states for typeahead input
Button shows loading of typeahead input selection.

## Screenshots
bvlpet15iu

_cc @noahmanger

### DIFF
--- a/js/typeahead-filter.js
+++ b/js/typeahead-filter.js
@@ -91,6 +91,8 @@ TypeaheadFilter.prototype.handleSelect = function(e, datum) {
     id: this.fieldName + '-' + slugify(datum.id) + '-checkbox'
   });
   this.datum = null;
+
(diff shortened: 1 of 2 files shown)


noahmanger:
👍 Only question I have is for @onezerojeremy / @jenniferthibault : 

Once a checkbox has been generated from the typeahead, should the loading indicator also appear next to the checkbox if you uncheck / recheck it? Right now it just shows up in the button the first time you add it, but then not again. It might be better as is, but wanted to check.

see

onezerojeremy:
Yes, I think unchecking a box should show a processing indicator each time, to let them know it's processing. not 100% sure how that works with the indicator being in the button though. Could the indicator just show up next to the check box, when you click it?

PR #456, opened 2016-08-05
remove bold from highlight

anthonygarvan:
Per discussion with @jenniferthibault.

### DIFF
--- a/scss/mixins/_type-mixins.scss
+++ b/scss/mixins/_type-mixins.scss
@@ -77,6 +77,5 @@
 
 @mixin t-highlight {
   background: $gray-light;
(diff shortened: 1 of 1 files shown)


jenniferthibault:
@anthonygarvan I went hunting to see where else this was being used, just to make sure, and found it in our typeahead searches. There the character bolding is a helpful distinction as you are live typing:

screen shot 2016-08-08 at 12 36 54 pm

Really sorry for doing this, and glad it was only 1 commit, but I'd like to not merge this in, and leave the bold styling.

noahmanger:
Closing, per @jenniferthibault

PR #460, opened 2016-08-11
Line chart styles & modals

noahmanger:
It's ugly, but this includes some new styles to support the [new chart work](https://github.com/18F/openFEC-web-app/pull/1508). 

- Activates nav links to "raising" and "spending" overviews, that are set to anchor links on /data
- Adds styles for line chart:
image
- Refactors `.overview__totals` as `.snapshot`, which accompanies the chart:
image
- Adds styles for modals:
image
- Note: modals require setting `main[aria-hidden="true"] { display: block }` because the modal library sets `aria-hidden="true"` on the `main` element when the modal is open, but we want it to still be visible behind the modal

### DIFF
--- a/js/templates/nav-data.hbs
+++ b/js/templates/nav-data.hbs
@@ -19,13 +19,13 @@
           </div>
         </form>
       </div>
-      <div class="mega__column is-disabled">
+      <div class="mega__column">
         <div class="mega-heading">
           <h3 class="mega-heading__title icon-heading--overviews">Interactive overviews</h3>
         </div>
         <ul>
-          <li class="mega__item is-disabled">Fundraising</li>
-          <li class="mega__item is-disabled">Spending</li>
+          <li class="mega__item"><a href="{{webAPpUrl}}#raising">Raising</a></li>
+          <li class="mega__item"><a href="{{webAPpUrl}}#spending">Spending</a></li>
         </ul>
       </div>
       <div class="mega__column">
--- a/js/templates/nav-data.hbs
+++ b/js/templates/nav-data.hbs
@@ -24,8 +24,8 @@
           <h3 class="mega-heading__title icon-heading--overviews">Interactive overviews</h3>
         </div>
         <ul>
-          <li class="mega__item"><a href="{{webAPpUrl}}#raising">Raising</a></li>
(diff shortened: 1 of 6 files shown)


noahmanger:
Ready for review. Updated with better description.

js/templates/nav-data.hbs
```diff
@@ -19,13 +19,13 @@
...
-      <div class="mega__column is-disabled">
+      <div class="mega__column">
         <div class="mega-heading">
           <h3 class="mega-heading__title icon-heading--overviews">Interactive overviews</h3>
         </div>
         <ul>
-          <li class="mega__item is-disabled">Fundraising</li>
-          <li class="mega__item is-disabled">Spending</li>
+          <li class="mega__item"><a href="{{webAPpUrl}}#raising">Raising</a></li>
+          <li class="mega__item"><a href="{{webAPpUrl}}#spending">Spending</a></li>
```
adborden: Typo here, `webAppUrl`

adborden: hmm, or should those be there at all? Maybe these should just be fragment urls?

adborden:
This looks beautiful. Is there a partner PR that goes with this, and has @jenniferthibault taken a look?

noahmanger:
Thanks! Yep, we paired up on it. PR over here: https://github.com/18F/openFEC-web-app/pull/1508

Change coming to those comments asap.

noahmanger:
Ok! Fixed the `webAppUrl` name and the modal:
image

scss/components/_modals.scss
```diff
@@ -0,0 +1,41 @@
+// Modal
+
+.modal-overlay {
```
xtine: Is this overriding a class set in the modal library or is there a reason why it isn't double dash like `.modal--overlay` to keep with convention?

noahmanger: Oh nope. That was just a mistake.

noahmanger: Fixed

adborden:
@xtine would you mind testing the modal for the issue you spotted and merge this once you're :+1: ?

xtine:
Modals are solid! 👍

PR #461, opened 2016-08-11
Making sidebar text color always white

jenniferthibault:
## Summary

Sets the sidebar title text color to `.inverse` so that it's white text over the primary or secondary background color at all times. This makes it just a bit easier on the eyes and more consistent across the site.

## Screenshots

**Before**
screen shot 2016-08-11 at 9 58 08 am

**After**
screen shot 2016-08-11 at 10 16 57 am

cc @adborden for https://github.com/18F/openFEC-web-app/pull/1505

### DIFF
--- a/scss/components/_sidebar.scss
+++ b/scss/components/_sidebar.scss
@@ -52,7 +52,7 @@
   margin: 0;
   padding: u(1rem 2rem);
   text-transform: uppercase;
-  @include u-font-color($inverse);
+  font-color: $inverse;
 }
 
 .sidebar__content {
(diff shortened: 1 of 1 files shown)


scss/components/_sidebar.scss
```diff
@@ -52,6 +52,7 @@
   margin: 0;
   padding: u(1rem 2rem);
   text-transform: uppercase;
+  @include u-font-color($inverse);
```
noahmanger: We don't need to use the mixin here. You can just do: `font-color: $inverse`

scss/components/_sidebar.scss
```diff
@@ -52,6 +52,7 @@
   margin: 0;
   padding: u(1rem 2rem);
   text-transform: uppercase;
+  font-color: $inverse;
```
noahmanger: Derp. And by `font-color` I meant `color`. Sorry!

PR #464, opened 2016-08-11
Calendar panel tweaks

noahmanger:
A few small changes to support https://github.com/18F/fec-cms/pull/424
- On small screens, set the `top` position of the filter panel with JS based on the position and size of the toggle button, rather than with CSS. This allows for variable content above the filter panel, as happens with the calendar.
- Adding a `emptyText` option to be passed to `FilterTags` so that the calendar can say `Viewing all events` when no tags are applied
- Minor CSS changes to adjust to new ways the calendar JS manipulates the DOM

### DIFF
--- a/js/filter-panel.js
+++ b/js/filter-panel.js
@@ -37,7 +37,7 @@ function FilterPanel(options) {
 }
 
 FilterPanel.prototype.setInitialDisplay = function() {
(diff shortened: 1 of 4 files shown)


noahmanger:
Can you take a look @adborden when you have a chance?

PR #465, opened 2016-08-12
Updated styles for advisory opinion pages

adborden:
### Data table heading borders

Adds a subclass `.data-table--heading-borders` to get cell borders in the header row. Some tables aren't adjacent to heading or a `.results-info` and thus don't have a top border, so having borders in the header row is awkward.

screenshot from 2016-08-11 18-11-36

### Heading styles

Create a new component based on the heading styles with a subtitle. I also moved the float property from the default `heading__title` to a new subclass `heading__title--with-action`.

screenshot from 2016-08-11 17-02-50

_cc @noahmanger @jenniferthibault _

### DIFF
--- a/scss/mixins/_type-mixins.scss
+++ b/scss/mixins/_type-mixins.scss
@@ -80,7 +80,3 @@
   font-weight: bold;
   padding: 0 0.3em;
 }
-
-@mixin t-upper {
-  text-transform: uppercase;
-}
--- a/scss/mixins/_type-mixins.scss
+++ b/scss/mixins/_type-mixins.scss
@@ -13,7 +13,7 @@
     padding: u(1.2rem 0);
 
     @include media($med) {
-      font-size: u(4.8rem);
+      font-size: u(4rem);
     }
   }
 
@@ -79,4 +79,8 @@
   background: $gray-light;
   font-weight: bold;
   padding: 0 0.3em;
+}
+
+@mixin t-upper {
+  text-transform: uppercase;
 }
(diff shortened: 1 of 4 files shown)


scss/mixins/_type-mixins.scss
```diff
@@ -80,3 +80,7 @@
   font-weight: bold;
   padding: 0 0.3em;
 }
+
+@mixin t-upper {
```
noahmanger: Just curious: why make a mixin for just one property?

noahmanger:
Looks good to me except for the one question about the mixin. 

Do you want to also add the padding fixes I mentioned on the other PR here? Or I can follow up with a separate one.

scss/mixins/_type-mixins.scss
adborden: I was thinking I would add documentation for it (but there are none for the `t-` mixins) and was the under the impression that was the norm. Was also unsure about the ettiquette using `@extend` for classes; some folks don't like `@extend`.

Anyway, I'll remove it, and while I'm at it, how about I do the same with `t-highlight`, and replace the `@include` usages with `@extend`?

adborden: Oh, I mis-understood your question... feeling really silly now :D Yeah, using `@include` or `@extend` seems pretty silly for a single property, especially one like "upper" that would never really having any other styles besides `text-transform: uppercase`.

I'll remove it and use the property directly. I won't touch `t-hightlight`.

noahmanger: 👍  all good. It's a fair question about `@extend` / `@include` . I don't have super strong thoughts on the matter other than that I know people tend to frown on `@extend`. But certainly not opposed outright.

PR #469, opened 2016-08-12
Renaming class to fix padding

noahmanger:
Fixes padding confusion.

On data table pages:
image

On legal search pages:
image


adborden:
Thanks!

PR #471, opened 2016-08-15
Beginning work on table toggle filter

noahmanger:
Creates a new `Filter` class for `ToggleFilter`, which consist of radio buttons. Also will host the disabling and enabling filter logic to go with https://github.com/18F/openFEC-web-app/pull/1522

### DIFF
--- a/js/filters.js
+++ b/js/filters.js
@@ -236,6 +236,31 @@ Filter.prototype.setLastAction = function(e, opts) {
   }
 };
 
+Filter.prototype.disable = function() {
+  this.$body.find('input, label, button').each(function() {
+    var $this = $(this);
+    $this.addClass('is-disabled').prop('disabled', true);
+    // Disable the tag if it's checked
+    if ($this.is(':checked')) {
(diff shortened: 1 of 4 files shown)


js/filters.js
```diff
@@ -234,6 +236,31 @@ Filter.prototype.setLastAction = function(e, opts) {
...
+
+Filter.prototype.enable = function() {
+  this.$body.find('input, label, button, .label').each(function() {
+    var $this = $(this);
+    $this.removeClass('is-disabled').prop('disabled', false);
+    $this.trigger('filter:enabled', {
+        key: $this.attr('id')
+      });      
+  });
+  this.isEnabled = true;
```
adborden: This is kind of minor, but I feel like these state changes should be at the top of the method, specifically, before the event trigger. The reason is that if someone hooked up to the event and inspected the state, it's a little strange that the filters `isEnabled` was false, even though we just saw a `filter:enabled` event.

Does that make sense?

noahmanger: Makes sense to me. Good call.

PR #473, opened 2016-08-16
New tag styles

xtine:
New design and functionality of tags:
- Tags are now shown in category blocks and separated by "or" text between them.

screen shot 2016-08-16 at 12 35 38 pm

--

screen shot 2016-08-16 at 12 35 59 pm

--

screen shot 2016-08-16 at 12 37 50 pm

--

screen shot 2016-08-16 at 12 35 19 pm

18F/openFEC-web-app#1380

### DIFF
--- a/js/dropdowns.js
+++ b/js/dropdowns.js
@@ -41,7 +41,6 @@ function Dropdown(selector, opts) {
     }
   }
 
-  $(document.body).on('tag:removed', this.handleRemoveClick.bind(this));
   $(document.body).on('tag:removeAll', this.handleClearFilters.bind(this));
 
   this.$button.on('click', this.toggle.bind(this));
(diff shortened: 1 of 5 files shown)


js/dropdowns.js
```diff
@@ -41,7 +41,6 @@ function Dropdown(selector, opts) {
     }
   }
 
-  $(document.body).on('tag:removed', this.handleRemoveClick.bind(this));
```
noahmanger: What did this do?

noahmanger:
Looks excellent. One thing, and I'm not sure if this introduced it or not, but when do the "clear all filters" button, on localhost:3000/receipts (for example), it removes the transaction period filter, which it shouldn't. Any `nonremovable` filter should stay applied after that.

js/dropdowns.js
xtine: I don't remember what the original purpose was, but with changing the functionality of tags this resulted in a bug that removed the checkbox entirely when a tag was removed.

xtine:
Good catch with the non-removable tag, just fixed that. I'll also write a test that sets a non-removable tag and then tests that it can't be removed with Clear all filters because I think nothing is checking for that yet.

jenniferthibault:
Just wanted to say WOW!!!
Can't wait to see how these test out.

PR #480, opened 2016-08-18
Better mobile layout for legal search

adborden:
## Summary

Part of the work for legal statutes https://github.com/18F/fec-eregs/issues/210.  Uses `display: table-cell` to switch between a tabular layout on larger screens and a compressed layout on mobile, taking into consideration some of the pattern discussion from https://github.com/18F/fec-style/issues/318.

@noahmanger let me know what you think. Wondering if it makes sense to make the `display: table` styles more generic, like using classnames like `.table`, `.thead`, `.td`, etc.

You can see this in the web app https://github.com/18F/openFEC-web-app/pull/1531

## Screenshots

**Before**
screenshot from 2016-08-18 12-59-13

**After**
screenshot from 2016-08-18 13-00-23

### DIFF
--- a/scss/components/_legal-search.scss
+++ b/scss/components/_legal-search.scss
@@ -86,6 +86,7 @@
     border-collapse: collapse;
     display: table;
     table-layout: fixed;
+    width: 100%;
   }
 
   .legal-search-results__header-row {
--- a/scss/components/_legal-search.scss
+++ b/scss/components/_legal-search.scss
@@ -33,38 +33,8 @@
 // Styleguide components.search-results-legal
 
 
-.legal-search-results__header-row {
-  border-bottom: 1px solid $primary;
-  font-family: $sans-serif;
-  font-weight: 600;
-
-}
-
-.legal-search-results__header {
-  display: none; // Hide all but the first header cell
-  padding: u(0.5rem 1rem);
-  text-align: left;
-
-  &:first-child {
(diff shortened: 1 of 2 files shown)


adborden:
ping @noahmanger :)

noahmanger:
Gah sorry. This looks nice. I like the idea of abstracting the styles though. Could we try to separate the responsive logic from the visual presentation (borders, colors, etc.), and make special classes for the former (`.table--responsive` or something) and use existing styles for the latter? 

`_table-styles.scss` could be a good home.

adborden:
:+1: So I moved most of the styles into `table-styles.scss` and turned them into subclasses `.simple-table--display` to handle the `.datatables`-esque styles and `.simple-table--responsive` for the responsive part. A few callouts:
- I had to move the bottom-border from cell to the row. I think this is cleaner, but wasn't sure if there was a reason for it to exist on the cell.
- There's a little hack around the padding. For small screens, I need to set the cell padding to `0`, so cells stack tightly, but I don't know the original padding for the media-query because it's based on the subclass you're using. So I added a special case for when you're using both `--display` and `--responsive`. You can see it when when using `--responsive` alone:
screenshot from 2016-08-23 16-26-41

adborden:
ping @noahmanger

noahmanger:
Nice! I like it.

PR #481, opened 2016-08-22
Fix election filter

noahmanger:
## Summary
Fixes the `ElectionFilter` so that it only fires the `filter:added` event first, and then `filter:renamed` after. This prevents it from generating multiple tags.

Resolves https://github.com/18F/openFEC-web-app/issues/1380#issuecomment-241424660

## Screenshots
demo

cc @xtine


xtine:
Nice, fix works. Thanks. 👍

PR #482, opened 2016-08-22
Breaking out filters.js into separate files

noahmanger:
## Summary
As a precursor to the new text filters I wanted to break out `filters.js` into separate files for each of the components, as it was getting unwieldy. I know that @xtine is working on the date filter though, so wanted to offer this up to get merged in sooner than later.

cc @xtine


xtine:
👍

Issue #458, opened 2016-08-09
Implement new financial range filters


noahmanger:
In order to make financial range filters easy to use and read, implement the new financial range filter design from https://github.com/18F/openFEC-web-app/issues/1493

![image](https://cloud.githubusercontent.com/assets/11636908/17502981/3547dc88-5dba-11e6-8fe3-58bddde120fc.png)

**Criteria for completion:**
- [x] Min and max filters display side-by-side, with a single button and hyphen in between
- [x] Text input is right aligned
- [x] The $ is always visible as a background image
- [x] Submitting either of the inputs triggers the loading and success state in the button

jenniferthibault:
Corresponding tag designs are in https://github.com/18F/openFEC-web-app/issues/1542

noahmanger:
Resolved by https://github.com/18F/fec-style/pull/484 . Nice work!

noahmanger:
Because we're not acting on the changes to the tag design yet (and have that captured in another issue), I'm closing this.

PR #484, opened 2016-08-24
Financial range filters

xtine:
New inline side by side input styles for financial ranges.

screen shot 2016-08-24 at 3 48 26 pm

For: #458
Requires: 18F/openfec-web-app#1544

cc: @noahmanger

### DIFF
--- a/scss/components/_form-styles.scss
+++ b/scss/components/_form-styles.scss
@@ -230,3 +230,67 @@
     border: 1px solid $error;
   }
 }
+
+
+// Financial range inputs
+//
+// Half width inputs so min and max filters display side by side
+//
+
+.range .js-filter {
+  display: inline-block;
+}
+
+.financial-range-input {
+  font-family: $sans-serif;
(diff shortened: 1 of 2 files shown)


scss/components/_form-styles.scss
```diff
@@ -230,3 +230,75 @@
...
+// Inline range inputs
+//
+// Half width inputs so min and max filters display side by side
+//
+
+.range {
+  margin-bottom: u(1rem);
+  position: relative;
+
+  .js-filter {
```
noahmanger: Can we use a different class here, so we're not using `.js-*` classes for styles?

scss/components/_form-styles.scss
```diff
@@ -230,3 +230,75 @@
...
+    display: inline-block;
+  }
+
+  .button--standard {
+    height: 3.6rem;
```
noahmanger: Need the `u()` here.

PR #486, opened 2016-08-25
Financial range tag

xtine:
Improved financial range tags, for 18F/openfec-web-app#1542.
- Setting one financial tag remains the same
- Setting an additional financial tag shows changed tag design with range inputs separated by a dash and in sequential order.
- Removing a financial tag in a range turns tag back to original singular form.
- Tags update in place

Also clicking enter to change a financial input will change focus to button like existing text input behavior.

7bwrrgllym

Requires 18F/openfec-web-app#1546

cc: @noahmanger

### DIFF
--- a/js/filter-base.js
+++ b/js/filter-base.js
@@ -126,7 +126,12 @@ Filter.prototype.handleChange = function(e) {
     loadedOnce = $input.data('loaded-once') || false;
 
     // set focus to button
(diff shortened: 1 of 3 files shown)


noahmanger:
Niiiiiice.

jenniferthibault:
WOOOO!!!!

PR #487, opened 2016-08-26
Document thumbnail component

adborden:
## Summary

Add a new document thumbnail component to visualize external documents as per [statutes landing page](https://github.com/18F/fec-eregs/issues/199#issuecomment-242517923). I didn't want to spend too much time tweaking this for small screens, so I set a `min-width` so when the screen gets very small, the text will stack under the thumbnail. The thumbnail is still left-aligned, so it's not perfect, but works.

@noahmanger also want to point out that I copied the styles from `.list-bulleted` into `.list-numbered` because the margins didn't seem right. I didn't see `.list-numbered` used anywhere, so figured this wasn't a big deal.

@jenniferthibault I guessed on the box-shadow, so feel free to tweak it or grab me and we can pair on it together.

## Screenshots

screenshot from 2016-08-25 17-15-25
--
screenshot from 2016-08-25 17-13-17


jenniferthibault:
taking a look now

jenniferthibault:
I added a commit to lighten up the box shadow just a tiny smidge!

I have a couple questions on the specific implementation on the statutes page:
- Can we crop the FEC doc (under option 1) to be about the same length as the other two? It looks like they exported it to a legal-sized paper, but it doesn't make sense to me to take up the page with an empty image

 
screen shot 2016-08-26 at 11 48 03 am


- It looks like the document thumbnails are wider than the text container for the rest of the statutes content. They should be the same width. 

 screen shot 2016-08-26 at 11 45 44 am

jenniferthibault:
I got started pretty early today, so I'm about to head out. If those changes don't get worked in before the release that's ok, not a hard stop.

If they don't get in, I'd like to make them a follow-up task for the next sprint though!

adborden:
I'll upload a cropped thumbnail for the campaign doc.

For the text container, the container size actually goes to the full page, but the `p` tags have a max width, so they don't fill the full container. We can trim the document grid size to match the `p` tags, but it didn't feel quite right to me.

ccostino:
Thank you, @jenniferthibault and @adborden!

Issue #491, opened 2016-08-29
ToggleFilter not adding tag


noahmanger:
I just noticed that the `ToggleFilter` (used for efiling vs processed filtering) isn't adding the associated tag: 

![image](https://cloud.githubusercontent.com/assets/1696495/18070853/fc5f7912-6e0d-11e6-8e54-30f452f95ced.png)
ug

PR #492, opened 2016-08-30
Ensure toggle filters add tags

noahmanger:
Fixes a bug where the toggle filters weren't triggering `filter:added` events. This fixes that bug and adds test coverage.

Resolves https://github.com/18F/fec-style/issues/491


xtine:
Nice test coverage 😎👍

PR #494, opened 2016-09-01
Improved Date Filter Interactions

xtine:
New date filter interactions to support a date grid picker.
18F/openfec-web-app#1364

screen shot 2016-09-01 at 3 19 26 pm
ditcqmdess

 
Date range tags have also changed to a new design, and now non-removable.
18F/openfec-web-app#1541

screen shot 2016-09-01 at 3 20 59 pm

Relies on 18F/openfec-web-app#1572


noahmanger:
This is working so well. I'm going to start logging some interaction thoughts I have:

The transaction period shouldn't have it's own tag:
image

There shouldn't be the "Receipt date" label between the time period filter and the "beginning" "ending" labels: 
Designed:
image

Actual:
image

It looks like it's missing this step in the interaction, moving focus to ending date after setting a beginning date:
image

And last thought: I want to workshop the timing / behavior of the way it disappears. You have it as designed as far as I can tell, but it feels a little jarring to me.

noahmanger:
Another thing: 

demo

I think the cursor should stay a pointer and the hover state should stay the same even after selecting a month for the first time. (Notice how it goes away after you've changed it once?)

scss/components/_date-grid.scss
```diff
@@ -0,0 +1,80 @@
...
+// Filter interaction for filter date fields
+//
+
+.date-range-grid {
+  @include clearfix();
+  font-family: $sans-serif;
+  padding: u(2rem 0);
+  display: none;
+
+  &:not(.invalid) {
```
noahmanger: Is the class `.invalid` or `.is-invalid`? If it's not the latter, should it be?

js/date-filter.js
```diff
@@ -20,11 +22,17 @@ function DateFilter(elm) {
...
+  this.$submit = this.$elm.find('button');
 
   this.$input.on('change', this.handleInputChange.bind(this));
   this.$elm.on('change', this.handleRadioChange.bind(this));
   this.fields = ['min_' + this.name, 'max_' + this.name];
 
+  this.$minDate.on('click', this.handleMinDateSelect.bind(this));
+  this.$maxDate.on('click', this.handleMaxDateSelect.bind(this));
+
+  this.$grid.find('li').on('click', this.dateGridSelect.bind(this));
```
noahmanger: What about `this.$elm.on('click', '.date-range-grid li'...)` instead?

js/date-filter.js
```diff
@@ -44,6 +52,8 @@ DateFilter.prototype.handleInputChange = function(e) {
   var $input = $(e.target);
   var value = $input.val();
   var loadedOnce = $input.data('loaded-once') || false;
+  var range = $input.data('range') || 'false';
```
noahmanger: Is there a reason `false` is a string instead of a boolean here?

js/date-filter.js
```diff
@@ -125,6 +146,127 @@ DateFilter.prototype.handleModifyEvent = function(e, opts) {
...
+DateFilter.prototype.setupDateGrid = function() {
+  var dateBegin = this.$minDate.val().split('/');
+  var dateEnd = this.$maxDate.val().split('/');
+  var dateRangeFirst = this.$grid.find('.date-range-first');
+  var dateRangeSecond = this.$grid.find('.date-range-second');
+  var minDateMonth = dateBegin[0];
+  var minDateYear = dateBegin[2];
+  var maxDateMonth = dateEnd[0];
+  var maxDateYear = dateEnd[2];
+  var transactionYear = $('#two-year-transaction-period').val();
```
noahmanger: This value is currently set as a property by the `handleModifyEvent` method, which is called when the transaction period changes. So I think you should be able to just reference it from the `this.maxYear` property.

js/date-filter.js
```diff
@@ -125,6 +146,127 @@ DateFilter.prototype.handleModifyEvent = function(e, opts) {
...
+  // the transaction year starts with the later, so begin before
+  dateRangeFirst.find('.date-range-year').html(parseInt(transactionYear - 1));
+  dateRangeFirst.find('ul').attr('data-year', parseInt(transactionYear - 1));
+
+  dateRangeSecond.find('.date-range-year').html(transactionYear);
+  dateRangeSecond.find('ul').attr('data-year', transactionYear);
+
+  // get the elements of the beginning and ending dates
+  $dateBegin = this.$grid.find('ul[data-year="' + minDateYear + '"] ' +
+    'li[data-month="' + minDateMonth + '"]').addClass('begin');
```
noahmanger: Can we make these `begin` and `end` classes more specific? Even just `.month--begin` / `.month--end`

js/date-filter.js
```diff
@@ -125,6 +146,127 @@ DateFilter.prototype.handleModifyEvent = function(e, opts) {
...
+  }
+};
+
+DateFilter.prototype.handleMinDateSelect = function() {
+  var self = this;
+  var $dateBegin = this.$grid.find('.begin');
+  var $dateEnd = this.$grid.find('.end');
+
+  this.$grid.show().removeClass('pick-max').addClass('pick-min');
+  this.$grid.find('.active').removeClass('active');
```
noahmanger: What about making this `.is-active`

js/date-filter.js
```diff
@@ -125,6 +146,127 @@ DateFilter.prototype.handleModifyEvent = function(e, opts) {
...
+  $dateBegin.addClass('active');
+
+  this.$grid.find('li').hover(
+    function() {
+      var dateBeginNum = parseInt($(this).parent().attr('data-year') + $(this).attr('data-month'));
+      var dateEndNum = parseInt($dateEnd.parent().attr('data-year') + $dateEnd.attr('data-month'));
+
+      if (dateBeginNum <= dateEndNum) {
+        self.$grid.removeClass('invalid');
+        self.handleDateGridRange($(this), $dateEnd, $dateBegin, $dateEnd);
```
noahmanger: Why are there four params passed here?

js/date-filter.js
```diff
@@ -125,6 +146,127 @@ DateFilter.prototype.handleModifyEvent = function(e, opts) {
...
+    value[1] = this.$maxDate.val();
+  }
+  else {
+    value[0] = this.$minDate.val();
+    value[1] = $selectDate.data('month') + '/01/' + $selectDate.parent().attr('data-year');
+  }
+
+  if (!this.$grid.hasClass('invalid')) {
+    this.$grid.removeClass('pick-min pick-max');
+    this.$grid.find('li').unbind('mouseenter mouseleave');
```
noahmanger: So this removes the event handlers added with the `$.hover` above?

scss/components/_date-grid.scss
```diff
@@ -0,0 +1,80 @@
...
+
+    &.pick-min li:hover div {
```
noahmanger: So the `.pick-min` and `.pick-max` classes are states applied to the whole grid depending on what's being picked? (Just making sure I understand).

scss/components/_date-grid.scss
```diff
@@ -0,0 +1,80 @@
...
+      border-top-left-radius: 15px;
+      border-bottom-left-radius: 15px;
```
noahmanger: Let's use `rem` just to be consistent.

js/date-filter.js
```diff
@@ -125,6 +146,127 @@ DateFilter.prototype.handleModifyEvent = function(e, opts) {
...
+DateFilter.prototype.handleMaxDateSelect = function() {
+  var self = this;
+  var $dateBegin = this.$grid.find('.begin');
+  var $dateEnd = this.$grid.find('.end');
+
+  this.$grid.show().removeClass('pick-min').addClass('pick-max');
+  this.$grid.find('.active').removeClass('active');
+  $dateEnd.addClass('active');
+
+  this.$grid.find('li').hover(
```
noahmanger: Do you think this will be an issue adding the event handlers here inside an event handler? I don't know what's best practice in this case.

js/date-filter.js
xtine: I don't think there was a particular reason, I can change it.

js/date-filter.js
xtine: Oh good catch, that was a leftover from when I was working on another version of the function that took 4 params but no longer does.

js/date-filter.js
xtine: Yeah, pretty much unbinding hover so it doesn't keep hover date changing after date select.

scss/components/_date-grid.scss
xtine: Yep, it's so the grid knows to hover/select the beginning or ending date.

js/date-filter.js
xtine: I can't think of any issues really, to me it just seems to encapsulate date hover within the date select click.

js/date-filter.js
xtine: Ahhh cool, that is handy.

scss/components/_date-grid.scss
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    &.selected div {
+      background-color: $gray-dark;
+      color: $inverse !important;
+    }
+
+    &.is-active div {
+      background-color: $aqua;
+      color: $base !important;
```
noahmanger: Is it possible to avoid the `!important`s?

xtine: Yes, thank you for the reminders. It was necessary when I was trying to change the color of just the `li` elements but since now the colors are set in a child `<div>` that has a specificity high enough not to need the importants.

js/date-filter.js
```diff
@@ -12,6 +12,8 @@ require('jquery.inputmask/dist/inputmask/inputmask.numeric.extensions.js');
 function DateFilter(elm) {
   Filter.Filter.call(this, elm);
   this.validateInput = this.$elm.data('validate') || false;
+  this.$range = this.$elm.find('.range');
+  this.$grid = this.$elm.find('.date__range--grid');
```
noahmanger: Ack, sorry, this is annoying, but can we tweak this to follow the BEM method more closely? `.block__element--modifier`. So in this case:

`.date-range__grid` where `.date-range` is the base block and `grid` is the sub-element, rather than a modifier.

Similarly: `.date-range__row`, `.date-range__year`, etc.

xtine: My bad, still getting used to BEM naming since I haven't used it before this project. Will change, thanks for the suggestion.

noahmanger:
I just added tests and made some small tweaks to the style and interaction of the filter in https://github.com/18F/fec-style/pull/494/commits/530faeed2344d32fc1e65518c2c55f37734568ca

Notably:
- After selecting a min date, focus now is set on the max date input, so you can just click on the max month 
- After selecting a max date, the focus is moved to the button
- I changed the `handleMinDateSelect` and `handleMaxDateSelect` handlers from `click` to `focus` to support this behavior
- Changed the formatting of dates to use slashes consistently
- Added some `.js-` specific classes for targeting DOM elements
- Renamed `dateGridSelect()` to `handleGridItemSelect()` to be more descriptive

@xtine if you're back and have time to review the tests, that'd be great. Otherwise maybe @ccostino?

xtine:
Tweaks and tests look good. This is seems ready to go, but I found an interaction bug that can be fixed after merge as well: #502

noahmanger:
Great. Yeah, that looks good. I made a couple small changes there to keep it in sync with my edits here. Will merge both once tests pass.

PR #497, opened 2016-09-06
Styles for new CMS features

noahmanger:
Because the `.rich-text` class is automatically applied to all blocks
from the CMS (with no way to remove it), this is required so that
nested lists display correctly.

---

## Update:
This pull request morphed as I did more work on the CMS. This now includes styles for the latest update posts and feed:

Post:
image

Feed filter:
image


noahmanger:
Updated the initial comment to explain what this is doing now that I committed some more changes.

scss/components/_tags.scss
```diff
@@ -18,6 +18,24 @@
...
+  border-radius: 4px;
+  color: $primary;
+  display: inline-block;
+  font-family: $sans-serif;
+  letter-spacing: -.3px;
+  padding: u(0 .5rem 0 .75rem);
+  margin: u(.5rem 3rem 0 0);
+}
+
+.tag--secondary {
```
ccostino: Was this meant to be a double underscore instead?

scss/modules/_hero.scss
```diff
@@ -69,6 +69,12 @@
   }
 }
 
+.hero--press {
```
ccostino: Same here - or is there a convention I'm missing?

scss/components/_tags.scss
noahmanger: Nope. The convention is `.base--modifier` and `.base__element` . So if it's just a variant, as in this case, of the same type of element, then it's double dash.

ccostino: Thanks!

ccostino:
Looks good, thanks @noahmanger!

PR #499, opened 2016-09-08
Adding file box icon

jenniferthibault:
## Summary

- Attempting to add the `file-box` icon (to be used for archived MURs) to the style guide
I've definitely added the file, but I think I'm missing steps to get it to appear in the style guide.

## Screenshots

screen shot 2016-09-08 at 11 26 40 am

cc @adborden : this is started, but any idea what I'm missing to get it to show up in the style guide list?


noahmanger:
This all good to go? I can merge.

jenniferthibault:
AHA I knew there was a command I was missing. Thanks @adborden. @noahmanger yep, if Aaron's addition completed adding the icon to the style guide in a way that makes it ready to use, then I think it's good to go.

adborden:
Ha! Not so fast! I'm not sure why, but the icon ends up like this. I thought maybe you had removed the fills since the mockup, but then looked at the source svg and realized something isn't getting processed right. Someone else want to give it a try?
screenshot from 2016-09-08 16-45-07

noahmanger:
Ah yeah same for me: 
image

This happens sometimes with the way the SVGs are built and exported. @jenniferthibault I forget how you've solved it in the past, but I think you know what to do?

jenniferthibault:
BLARGH!

Yes. This is caused by bugs in how Sketch subtracts vector layers. Sometimes everything looks perfect when it renders, sometimes things turn into one solid blob. Thanks for flagging! There's no way to tell if the icon is correct until the it renders for the first time, which for us is in the style guide. 

The fix for this is to build and export the icon in Illustrator, which I'll do now.

jenniferthibault:
@adborden @noahmanger I added the new file, but when I ran `npm run build-icons` it:
- generated a lot of changes to `_icon-variables.scss`
- but didn't show the icon at all in the style guide for me.

So I didn't commit `_icon-variables.scss` 

At some point in a slower sprint I'd love to have one of you pair and walk me through this, since I keep doing something wrong!

noahmanger:
Bingo:
image

The missing step for getting it to show up on the styleguide is to run `npm run build`. 

But we should troubleshoot why we're all getting different results from running `build-icons`.

adborden:
Awesome, thanks all!

PR #506, opened 2016-09-12
updating ethn.io survey ID

porta-antiporta:
## Summary

Updating ethnio screener to: https://ethn.io/70862

## Screenshots

- _Update the styleguide documentation, include a screenshot if applicable._
- _Include a screenshot of the new/updated styles in context (“in the wild”)._

_cc @username_

### DIFF
--- a/js/templates/feedback.hbs
+++ b/js/templates/feedback.hbs
@@ -14,7 +14,7 @@
     <form id="feedback-form" class="container">
       <fieldset>
         <legend class="feedback__title">Help us improve betaFEC</legend>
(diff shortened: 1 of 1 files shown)


noahmanger:
Just checking: where does the current ethnio screener link go that this is replacing? cc @onezerojeremy

onezerojeremy:
if i understand the question correctly, the old link goes to a deactivated screener (we were using it while we were testing the IA, it had a different prompt, and turned it off afterward). 

in the future, knowing that the screener is hard-coded to feedback box, we'll have to be more careful and only update the screener, not actually deactivate it.  (tagging @nickykrause @emileighoutlaw @ethanheppner so we all know)

noahmanger:
Cool cool.

PR #510, opened 2016-09-15
Feature/murs

adborden:
## Summary

Adding styles for the canonical MUR pages (for archived MURs). Part of https://github.com/18F/fec-eregs/issues/227

## Screenshots

screenshot from 2016-09-15 14-37-23

### DIFF
--- a/scss/components/_legal-search.scss
+++ b/scss/components/_legal-search.scss
@@ -45,14 +45,30 @@
   }
 }
 
-.legal-mur__archive {
-  font-size: u(1.3rem);
-  margin-top: u(0.8rem);
+.legal-mur {
+  .legal-mur__archive {
+    font-family: $sans-serif;
+    font-size: u(1.3rem);
+    margin-top: u(0.8rem);
 
-  .legal-mur__archive-icon {
-    @include u-icon($file-box, $base, 1.5rem, 1.5rem, 100%);
-    display: inline-block;
-    margin-right: u(0.7rem);
-    vertical-align: top;
+    .legal-mur__archive-icon {
+      @include u-icon($file-box, $base, 1.5rem, 1.5rem, 100%);
+      display: inline-block;
+      margin-right: u(0.7rem);
+      vertical-align: top;
+    }
+  }
+
+  .legal-mur__property {
+    margin-bottom: u(1rem);
+  }
+
+  .legal-mur__subject-list {
+    list-style: disc;
+    margin-left: u(2rem);
+  }
+
+  .legal-mur__subject-list-item {
+    margin-left: u(1rem);
   }
 }
--- a/scss/components/_legal-search.scss
+++ b/scss/components/_legal-search.scss
@@ -80,3 +80,22 @@
     margin-left: u(1rem);
   }
 }
+
+.legal-top {
+  font-family: $sans-serif;
+  border-bottom: none;
+
+  &::before {
+    @include u-icon-bg($arrow-up-circle, $gray);
+    content: '';
+    display: inline-block;
+    height: 2rem;
+    margin-right: 0.5rem;
+    vertical-align: top;
+    width: 2rem;
+  }
+
+  @include media($med) {
+    display: none;
+  }
+}
--- a/scss/components/_legal-search.scss
+++ b/scss/components/_legal-search.scss
@@ -49,7 +49,11 @@
   .legal-mur__archive {
     font-family: $sans-serif;
     font-size: u(1.3rem);
-    margin-top: u(0.8rem);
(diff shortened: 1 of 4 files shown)


scss/modules/_datatables.scss
```diff
@@ -115,10 +115,15 @@
     background-color: $inverse;
     border-bottom: 1px solid $neutral;
 
+    &.row-color-contrast,
     &:nth-child(even) {
       background-color: rgba($gray-lightest, .5);
     }
 
+    &.row-color-normal {
```
jenniferthibault: This is just for my own understanding: if line 115 sets the whole table background to `$inverse`, why do you need to set a separate `normal` row color to the same color?

adborden: Sure, this is so that you can override the `nth-child(even)` rule on line 119 which sets every other row to background gray. The main use case is when cells span multiple rows, like some of the tables in the legal pages. With the `nth-child` rule, you get something like this:

(This only works because github has the same `nth-child` rule for markdown tables)

<table>
<tbody>
<tr>
  <td rowspan="2">I span 2 rows</td>
  <td>one row</td>
</tr>
<tr style="background-color: #eee;">
  <td>one row</td>
</tr>
<tr>
  <td rowspan="3">I span 3 rows</td>
  <td>one row</td>
</tr>
<tr  style="background-color: #eee;">
  <td>one row</td>
</tr>
<tr>
  <td>one row</td>
</tr>
</tbody>
</table>

Markup:
```html
<table>
<tbody>
<tr>
  <td rowspan="2">I span 2 rows</td>
  <td>one row</td>
</tr>
<tr>
  <td>one row</td>
</tr>
<tr>
  <td rowspan="3">I span 3 rows</td>
  <td>one row</td>
</tr>
<tr>
  <td>one row</td>
</tr>
<tr>
  <td>one row</td>
</tr>
</tbody>
</table>
```

You can see how it's still highlighting the even rows, but that's not actually what we want. We really want it to consider the first column cell to be considered a row.

So to apply the new classes, the even rule still applies, unless you override it with `.row-color-normal`. So this markup would get you white-gray alternating at the row group level the way the legal mockups are (sorry, github won't let me demo this one).

Markup
```html
<table>
<tbody>
<tr class="row-color-normal">
  <td rowspan="2">I span 2 rows</td>
  <td>one row</td>
</tr>
<tr class="row-color-normal">
  <td>one row</td>
</tr>
<tr class="row-color-contrast">
  <td rowspan="3">I span 3 rows</td>
  <td>one row</td>
</tr>
<tr class="row-color-contrast">
  <td>one row</td>
</tr>
<tr class="row-color-contrast">
  <td>one row</td>
</tr>
</tbody>
</table>
```

Does that makes sense?

jenniferthibault: Yes!! thank you!

jenniferthibault:
Visual styles = 👍 , but I can't speak to technical review.

scss/components/_legal-search.scss
```diff
@@ -45,14 +45,30 @@
...
+      margin-right: u(0.7rem);
+      vertical-align: top;
+    }
+  }
+
+  .legal-mur__property {
+    margin-bottom: u(1rem);
+  }
+
+  .legal-mur__subject-list {
```
noahmanger: Could you just use the more abstract `.list--bulleted` class here?

adborden: [.list--bulleted](https://github.com/18F/fec-style/blob/master/scss/components/_list-styles.scss#L23) actually does more than just the margins, it also adds padding and font-size. I can clean that up though, maybe move that into a `.list--large` or `.list--spaced` ?

noahmanger: Ah hmmm. That might be harder to change because it's used a lot. That approach could work, or we could have that stay the same and a new `.list--dense` or something? Or just using your approach here works too. I don't have strong feelings.

adborden:
On Mon, 2016-09-19 at 17:08 -0700, Noah Manger <notifications@github.com> wrote:
> noahmanger commented on this pull request.
> 
> Ah hmmm. That might be harder to change because it's used a lot. That approach could work, or we could have that stay the same and a new `.list--dense` or something? Or just using your approach here works too. I don't have strong feelings.

Hmm, turns out there's already a `.list--spacious`. I'm inclined to
leave it as is. Not a fan of `.list--dense` only because it would be
overriding styles, which gets a little messy.


--
Aaron D Borden
Human and Hacker

adborden:
@noahmanger ack, forgot I had one more commit based on @jenniferthibault's feedback, just pushed it. Please merge if it's all good.

Issue #507, opened 2016-09-12
Date picker "end" date should select the last of the month


noahmanger:
@PaulClark2 raised a good point in our demo that I think should be implemented: When selecting an ending month in the new date picker, it currently sets the end date to the first of that month. It's probably more intuitive to have it set the last date in the month. 

Current behavior:
![image](https://cloud.githubusercontent.com/assets/1696495/18455085/b3214c72-78fc-11e6-9985-66fe4b87f076.png)

If anyone has an argument against this, I'd love to hear, because it seems pretty clearly a good move to me.

cc @nickykrause @jenniferthibault @xtine

nickykrause:
That's a great point, especially given the associated visual, wherein September appears "filled in" or "selected," which, at least to me, suggests that the date range _includes_ September (i.e., the selection goes to the end of the month). 

Changing this makes sense to me.

jenniferthibault:
yup!

noahmanger:
Resolved by #https://github.com/18F/fec-style/pull/511

cc @PaulClark2 : Thanks for the good suggestion!

PR #511, opened 2016-09-16
Date Filter - set last day of month on end

xtine:
Selecting an end date on the date filter picker will set the last day of the month. #507 

screen shot 2016-09-15 at 10 01 43 pm


noahmanger:
Nice!

PR #514, opened 2016-09-16
Fix grid for multiple rows

adborden:
## Summary

I was playing around with the grid, but it wasn't working as expected for 3 and 4 column grids. The main issue was that the `nth-child` margin rules were being applied multiple times, at different intervals because the breakpoints overlap. The issue is kind of hidden from the styleguide because we're not including enough grid items.

To fix this, I introduced a one-off media query that has a `max-width` to prevent the breakpoint overlap. This is an exception to how we usually define breakpoints, but I think the implementation is reasonable (at least we're using the `$large-screen` variable).

#### 2-wide
screenshot from 2016-09-16 10-36-45

#### 3-wide
screenshot from 2016-09-16 10-36-56

#### 4-wide
screenshot from 2016-09-16 10-37-06

#### 2-wide
screenshot from 2016-09-16 10-38-02

#### 3-wide
screenshot from 2016-09-16 10-38-09

#### 4-wide
screenshot from 2016-09-16 10-38-21

cc @noahmanger

### DIFF
--- a/scss/layout/_grid.scss
+++ b/scss/layout/_grid.scss
@@ -65,18 +65,15 @@
   }
 }
 
-// BREAKPOINT: MEDIUM
-@include media($med) {
+// BREAKPOINT: LARGE
+@include media($lg) {
   .grid--2-wide {
     .grid__item {
       @include span-columns(6);
       @include omega(2n);
     }
   }
-}
 
-// BREAKPOINT: LARGE
-@include media($lg) {
   .grid--3-wide {
     .grid__item {
       @include span-columns(4);
--- a/scss/layout/_grid.scss
+++ b/scss/layout/_grid.scss
@@ -20,12 +20,24 @@
 //   <div class="grid__item">
 //     Grid item
 //   </div>
+//   <div class="grid__item">
+//     Grid item
+//   </div>
+//   <div class="grid__item">
+//     Grid item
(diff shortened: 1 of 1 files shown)


noahmanger:
Oooooh nice. I'll need to take a closer look, but thanks for figuring this out!

scss/layout/_grid.scss
```diff
@@ -52,31 +65,29 @@
   }
 }
 
+// BREAKPOINT: MEDIUM
+@include media($med) {
```
noahmanger: Is this duplicative of the one above?

adborden: Kind of, the important difference is that this one applies to `$med` and up, while the previous one doesn't. Without this, the `grid--2-wide` becomes 1 column on large screens.

Not sure if there's a cleamer way to structure it, let me know if you have suggestions :)

noahmanger: Ah! Well if the above media query handles `$med` - `$lg` then should this one just be `$lg` and up?

adborden: Awesome! Obvious and much clearer!

noahmanger:
Is this ready to merge? Looks good to me, just noticed the label wasn't on so didn't want to merge prematurely.

PR #515, opened 2016-09-16
Cleaning up styles on lists and tables on custom pages

noahmanger:
## Summary
Makes several small changes to clean up / fix styling on custom pages, including the new press content pages.

## Screenshots
Creates a new class `.list--bulleted--alt` in which only the first level of `<li>` have bullets, and nested `<ul>` don't:
image

Restores standard bulleted list and table styling to the custom pages:

image

Cleans up unnecessary extra styles from the `.rich-text` class by just using `@extend`

Adds some special styles to account for the fact that scraped press releases contain tables for layout.

Resolves https://github.com/18F/fec-cms/issues/494

cc @emileighoutlaw @adborden

### DIFF
--- a/scss/components/_list-styles.scss
+++ b/scss/components/_list-styles.scss
@@ -2,10 +2,11 @@
 //
 // These are classes to be applied to lists to give them particular styles
 //
-// .list--bulleted  - Bulleted list (for ul)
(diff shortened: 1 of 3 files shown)


emileighoutlaw:
Woohoo! This is awesome. Thanks, Noah.

Issue #501, opened 2016-09-09
Update DateFilter component to work with non-transaction-period dates


noahmanger:
We want to implement the new date filter style across all date filter instances, including those that aren't bound by a two-year transaction period. The code currently requires that for setting min and max years, though, so it needs to be updated in order to finish applying it everywhere.

xtine:
Where are the instances of date filter that isn't bound by a two year transaction period?

noahmanger:
Basically any transaction or report page other than receipts and disbursements. For example:
- http://localhost:3000/independent-expenditures/
- http://localhost:3000/filings/
- http://localhost:3000/reports/presidential/

xtine:
Ah, is it the radio date selection? Is it just selecting date in the "custom" option or also within the cycle?

noahmanger:
We want to replace the radio button pattern and instead have a single date-picker interaction. The difference being that these won't be restricted to any given two-year period. 

I think it's fine for the months to default to the months in years we're actually in, though. Maybe if you enter a date that's earlier the months don't show?

I don't know that we completely fleshed out the interaction here, so tagging in @onezerojeremy and @nickykrause if they have thoughts.

jenniferthibault:
Recapping our convo today, we're going to start minimal, and use the same date text/entry boxes for Beginning and Ending fields, but remove the date picker interaction for non-cycle-bound instances altogether.

That looks like this:

![screen shot 2016-09-19 at 4 43 53 pm](https://cloud.githubusercontent.com/assets/11636908/18648386/6d94bee6-7e88-11e6-8b56-ab11a3aeaf0c.png)


TBD through testing is if users expect any graphic date selection aid to appear.

noahmanger:
Resolved by #517

PR #517, opened 2016-09-19
Simplify DateFilter for non two-year transaction periods

xtine:
Date filters that are not bound to two-year transaction periods will not have a radio selection but  simpler beginning/end input date fields. Relies on 18F/openfec-web-app#1590.

Issue: #501

screen shot 2016-09-19 at 3 04 38 pm

### DIFF
--- a/js/date-filter.js
+++ b/js/date-filter.js
@@ -25,11 +25,6 @@ function DateFilter(elm) {
     oncomplete: this.validate.bind(this)
   });
 
-  if (this.$minDate.attr('data-min-date')) {
-    this.$minDate.val(this.$minDate.data('min-date')).change();
-    this.$maxDate.val(this.$maxDate.data('max-date')).change();
-  }
-
   this.$input.on('change', this.handleInputChange.bind(this));
 
   this.fields = ['min_' + this.name, 'max_' + this.name];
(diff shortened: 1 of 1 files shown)


js/date-filter.js
```diff
@@ -25,8 +25,13 @@ function DateFilter(elm) {
     oncomplete: this.validate.bind(this)
   });
 
+  if (this.$minDate.attr('data-min-date')) {
```
noahmanger: Is this necessary? I don't think we should be setting min and max dates by default here. It's ok to have them empty.

xtine: Is it really okay to have the date input fields empty? The page first loads with a one year period by default, so I think the fields should reflect that.

noahmanger: It does? The nav link includes a `max_date` (which is there to hide expenditures that were incorrectly marked as happening in the future), but other than that I don't think we should be limiting by date automatically. (In principal I agree with you, though).

xtine: Ah okay so my radio button was set to "this year" so I had mistakenly thought that was default. I've taken out default values for date input fields.

xtine:
Date filter tags for non two-year transactional periods now removable:
screen shot 2016-09-20 at 10 25 09 am

noahmanger:
Nice! The only thing I caught is that it doesn't seem to de-increment the filter count when removing one of these tags. (The others seem to be working as expected).

xtine:
Had to set the non-transactional date inputs as loaded so the `handleRemoveEvent` in `filter-panel.js` decrements load count correctly.

Issue #488, opened 2016-08-26
Filters creates a right gutter on mobile


adborden:
There's a `.filters label::after` rule that creates a 20px gutter on the right side of the screen for mobile. Looks like it's for the loading animation but acting up on mobile.

<img alt="screenshot from 2016-08-25 18-36-22" src="https://cloud.githubusercontent.com/assets/509703/17991381/ece781e6-6af2-11e6-9971-4fb5cbc80ee8.png" width="300">

adborden:
Digging a little deeper, it's because the label on the search text field is so wide, having the `label::after` hangs off the page. setting `position: relative` on the `.filter` class helps, but there's still a bit of a gutter.

xtine:
@adborden: Is this still an issue? Tried to load this at mobile size and it looks fine to me:

<img width="330" alt="screen shot 2016-09-19 at 3 11 28 pm" src="https://cloud.githubusercontent.com/assets/24054/18650721/6c75286e-7e7b-11e6-9a5b-8d533d0e5536.png">

adborden:
Yep, still an issue https://fec-dev-proxy.18f.gov/data/candidates/?election_year=2012&election_year=2014&election_year=2016&has_raised_funds=true

Just scroll right. I'm on linux, but I see it in every browser, chrome, firefox, and epiphany (webkit)
![screenshot from 2016-09-19 17-13-40](https://cloud.githubusercontent.com/assets/509703/18653114/8d50dbee-7e8c-11e6-8c5d-1394d6faf63f.png)

adborden:
It also happens on my mobile, but only when the filters are open. If you still can't repro it I'm happy to pair with you on it.

noahmanger:
Yeah, I've seen this too. I think in iPhones it defaults to zooming you in so you don't notice it, but if you scroll over it's there.

xtine:
Ahh, I didn't see it because I didn't try to scroll right. Just pushed up a fix.

noahmanger:
Resolved by #518

PR #518, opened 2016-09-20
Fix filter gutter on mobile

xtine:
Tweaked the default negative right margins on filter labels so it doesn't create a gutter on mobile screens. Issue #488

### DIFF
--- a/scss/modules/_filters.scss
+++ b/scss/modules/_filters.scss
@@ -104,6 +104,7 @@ $filter-button-width: u(4.6rem);
       &::after {
         @include u-icon-bg($x, $error);
         background-position: 50%;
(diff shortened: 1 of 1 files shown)


scss/modules/_filters.scss
```diff
@@ -94,6 +94,7 @@ $filter-button-width: u(4.6rem);
       &::after {
         @include u-icon-bg($check, $green-light);
         background-position: 50%;
+        right: u(-3rem);
```
noahmanger: Do we need this on `.is-unsuccessful` as well?

xtine: Ah good catch. Just added it.

Issue #490, opened 2016-08-29
Reconcile h1 main headings


jenniferthibault:
Based on https://github.com/18F/fec-style/issues/467 , reduce the number of classes it takes to build a main header, and implement them consistently across templates. 

(Holding for final designs)

noahmanger:
@jenniferthibault ran into a little snag. If all `headings` have 30px margin at the bottom, how should the press release / digest heading pattern work. Here's what it does by default with the new styles:

![image](https://cloud.githubusercontent.com/assets/1696495/18730473/5b61ebe8-800b-11e6-9aab-cca3cb630e67.png)

jenniferthibault:
@noahmanger is the problem with that screenshot that the button is creating additional clearance?

Sorry, if that's not what you mean, then I'm not really sure what the snag is.

noahmanger:
Oh sorry. Yeah, so basically, by default, our `heading` class has 30px space below. Which when applied in this case forces extra space between the border and the button. So wondering if the design rule is "30px space between the heading and everything below it" or "count the button as part of the heading and the 30px should be below the button".

noahmanger:
Bumping this again now that the release is out. @adborden do you want to take a look and see if there will be implications for any work you did between when I started this and now?

jenniferthibault:
I'm diving in now to look around more thoroughly, but in general I think the rule is that the main text column and right-hand columns should align evenly at the top. In this case, the button should be in the right-hand column, not in a row above the body copy. It should look aligned, like the image in your CMS work:
https://github.com/18F/fec-cms/pull/509

noahmanger:
Ah great! So it's in the "sidebar" column as part of the body, not part of the header?

jenniferthibault:
Yes! Thanks for helping me understand.

adborden:
@noahmanger there are, I have two PRs up, but I will update them.

PR #520, opened 2016-09-21
Cleaning up heading styles

noahmanger:
## Summary
This refactors our heading Sass to clean up cruft and simplify and unify the styles for page and section headings. Because headings can contain multiple elements — a title (as an `h1`, `h2`, etc.), and optionally a subtitle, an action, or a tag — instead of applying the border style to the `h*` elements, I made two new classes: `.heading--main` and `.heading--section`, which are applied to a `div` that wraps the whole set of elements. (I'm open to renaming these classes if there's other names that would be clearer). 

So the new structure for a page header, for example, is:
```
<div class="heading--main">
  <h1>Page heading</h1>
</div>
```

The stuff inside only needs a class if it's either a subtitle or if there's an action (buttons, selects, etc.) that needs to be floated. In that case, we just use `.heading__left` on the title and `.heading__right` on the button or whatever action.

Refactoring something like this is definitely tricky and it's easy to miss things, so I'm going to ask @jenniferthibault to take a careful look through. 

## Screenshots
In the styleguide:
image

In the wild:
image

image

image

image

cc @jenniferthibault @adborden 

Resolves https://github.com/18F/fec-style/issues/490
Requires https://github.com/18F/openFEC-web-app/pull/1592
Requires https://github.com/18F/fec-cms/pull/509


scss/components/_headings.scss
```diff
@@ -0,0 +1,87 @@
...
+//  </div>
+//
+// Styleguide components.headings-with-action.1
+
+.heading--with-action {
+  h1,
+  h2,
+  h3,
+  h4 {
+    line-height: u(3.4rem);
```
adborden: Is this a well-known font-size or something? I thought it was best practice [not to specify units](https://pages.18f.gov/frontend/#measurements) for `line-height`.

adborden:
:+1: awesome! I really like the additional documentation.

scss/components/_headings.scss
noahmanger: Yeah, I agree not best-practice. This is kind of a hack in order to have the buttons and heading line up correctly, like:


Not sure if there's a better way

jenniferthibault:
@noahmanger this. is awesome. I am so excited. I left a few comments above (mostly for my own clarity). The only thing I'm seeing as a possible error is a little funkiness on the Headings with dates style:

screen shot 2016-10-04 at 5 29 17 pm

scss/components/_headings.scss
```diff
@@ -0,0 +1,124 @@
+// Headings
+//
+// Contain the title, date select and action button.
+// Use .heading--main when the heading is the top-level heading on a page as it provides extra margin-bottom
+// Use .heading--section When the heading is just a heading of a section on a page for less margin-bottom
```
noahmanger: Just figured out that we can use markdown formatting here to format, so I used the tic marks.

scss/components/_headings.scss
```diff
@@ -0,0 +1,124 @@
...
+  letter-spacing: -.3px;
+  margin-bottom: u(1rem);
+  text-transform: uppercase;
+}
+
+// Headings with actions
+//
+// When the heading title includes an action, add this class to one of the base classes.
+// Set line height to the height of the buttons or selects in the heading
+// Use .heading__left and .heading__right to float elements at $med breakpoints
```
noahmanger: Fixed when I figured out it's because of markdown formatting.

noahmanger:
All good!

scss/components/_headings.scss
adborden: :+1: let's just add a comment that the 3.4rem is coming from the size of the action button.

Issue #449, opened 2016-08-03
Fix calendar filter counts


ccostino:
The calendar filters were susceptible to an issue that is being temporarily fixed by #448, but we want to look into this further and fix it more appropriately.  From the PR:

> This is a temporary fix to prevent a bug from shipping out, but we'll want to actually fix this eventually. Here's the problem: the calendar filters each have the same name attribute: category. So the check that we have on the handleAddEvent method to see see if the event name === the filter name passes for all of them on every filter:added event.

> The reason they all share the same name is that we really just have one API filter for category but on the frontend we break it out by different subgroups. That's what the MultiFilter class is designed to handle.

noahmanger:
Resolved by https://github.com/18F/fec-style/pull/522

PR #522, opened 2016-09-22
Fix calendar filter counts

xtine:
The issue (#449) with inflated calendar filter counts is caused by multiple filters sharing the same name. This can be remedied simply by setting the `this.name` variable to the unique label instead, thus preventing inflated filter counts.

screen shot 2016-09-22 at 12 58 35 pm screen shot 2016-09-22 at 1 00 08 pm

cc @noahmanger


noahmanger:
Nice!

PR #523, opened 2016-09-27
Add styles for enforcement landing

adborden:
## Summary

Includes styles for the enforcement matters landing page. Introduces a `.slab--inline` for non-full-width slabs, like the MURs search. Also creates a `.u-no-print` utility to hide elements for print.

## Screenshots

screenshot from 2016-09-26 19-27-45

### DIFF
--- a/scss/components/_search-bar.scss
+++ b/scss/components/_search-bar.scss
@@ -42,11 +42,6 @@ $search-button-width: u(5.6rem);
 
 .search__example {
   margin-top: u(1rem);
-
-  // For aligning the example text after a select input
-  &.search__example--with-select {
-    margin: u(1rem) 0 0 u(21rem);
-  }
 }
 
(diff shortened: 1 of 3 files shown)


scss/components/_search-bar.scss
```diff
@@ -42,6 +42,17 @@ $search-button-width: u(5.6rem);
 
 .search__example {
   margin-top: u(1rem);
+
+  // For aligning the example text after a select input
+  &.search__example--with-select {
+    margin: u(1rem) 0 0 u(21rem);
```
noahmanger: Do you want to set this `margin-left` on small screens by default or should it only be on larger screens?

adborden: Yep, good call. Should only be medium and up like it is today. Fixed.

PR #524, opened 2016-09-27
Footer updates: new social media icons and styling

xtine:
In support of 18F/openFEC-web-app#1608, new social media icons and styles for footer updates.


xtine:
I'm not really fussed with icon data changing with compilation by other users, as long as a build doesn't break I don't think it's an issue to worry about.

PR #526, opened 2016-09-28
Changing font size

noahmanger:
Bumps the font-size down of the snapshot controls for the data landing page charts.

This is a somewhat hacky solution to prevent awkward breaking of the component at certain window widths. 

I'm going to keep https://github.com/18F/fec-style/issues/525 open as an issue to come up with a more resilient, less brittle solution in the future.

cc @jenniferthibault


ccostino:
Thanks, @noahmanger!

PR #530, opened 2016-09-30
Enable support for custom tag values in data attributes

noahmanger:
This enables passing custom tag values in `ToggleFilter` filters, which is useful in the case of when a filter is simply a boolean. A tag that just says `true` or `false` isn't much use, so this allows you to set a `data-tag-value` on the input and that will be used as the value for the tag. 

Goes with https://github.com/18F/openFEC-web-app/pull/1565

### DIFF
--- a/js/toggle-filter.js
+++ b/js/toggle-filter.js
@@ -20,12 +20,13 @@ ToggleFilter.prototype.fromQuery = function(query) {
 };
 
 ToggleFilter.prototype.handleChange = function(e) {
+  var $input = $(e.target);
   var value = $(e.target).data('tag-value');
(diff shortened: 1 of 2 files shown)


ccostino:
+1 to @xtine's remark.  What if you had the `data-tag-value` and a `data-tag-label` to compliment it to account for things like the `Data type` string?

noahmanger:
Good call. Made the change to use `data-prefix`, which requires a tweak over here: https://github.com/18F/openFEC-web-app/pull/1630

ccostino:
Awesome, thanks @noahmanger!

PR #531, opened 2016-10-03
Updating press image

jenniferthibault:
## Summary

- Updates the press image.
  - (Note: we don't love that there isn't good inclusion of gender in this image. Our ideal choice would have better representation of women and non-white men.)
- Attempts to update the style guide (but could use some final-stretch help from @noahmanger with the following issues) 
  - The "Show markup" isn't displaying as a button, and is instead replacing the "Examples" text. 
  - The fonts are using the style guide fonts, not the site fonts for some reason
  - Is it possible to get the photos to show up here?

### In the wild

screen shot 2016-10-03 at 10 49 45 am

### Style guide

screen shot 2016-10-03 at 12 21 09 pm
screen shot 2016-10-03 at 12 21 16 pm

## Review:
@noahmanger


noahmanger:
I got it working: 
image

Not seeing the issue with the fonts. The issue with the images was solved by deleting the `/styleguide` folder and running `npm run build` again. 

Also, I moved the example markup to a separate `heroes.hbs` file to keep this file tidier since it was really long.

jenniferthibault:
Thank you!!! And neat to know about moving files to `.hbs` folders when they get long. Very excited about this.

PR #533, opened 2016-10-04
Alert consistency

emileighoutlaw:
On our receipts page, the alert message for selecting a date that falls outside the two-year period currently says

"You entered a date that's outside the selected transaction period." But there's no _thing_ on the page called "transaction period," which is confusing.  

Screenshot:
screen shot 2016-10-04 at 4 03 53 pm

This PR is to make the language more consistent with the filters we already have applied. 

cc @jenniferthibault @noahmanger @xtine


jenniferthibault:
Nice spotting!

PR #545, opened 2016-10-17
Filings/Reports table document download link

xtine:
Styles for the document download link on Filings and Reports tables. Supports 18F/openFEC-web-app#1635

### DIFF
--- a/scss/modules/_data-container.scss
+++ b/scss/modules/_data-container.scss
@@ -84,6 +84,16 @@
   .dropdown {
     margin-top: u(.5rem);
   }
+
+  .dropdown__panel {
+    min-width: 70px;
+    right: u(1rem);
+    margin-top: u(.15rem);
+  }
+
+  .dropdown__value {
+    padding: u(0 .75rem);
+  }
 }
 
 @include media($med) {
--- a/scss/modules/_data-container.scss
+++ b/scss/modules/_data-container.scss
@@ -42,7 +42,6 @@
   background-color: $secondary;
 }
 
-
 .data-container__head {
   @include clearfix();
   border-bottom: 1px solid $gray;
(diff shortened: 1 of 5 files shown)


jenniferthibault:
Oops, I meant to put this comment about the button style in THIS pr, not the one I did:
https://github.com/18F/openFEC-web-app/pull/1635#issuecomment-254329472

scss/modules/_data-container.scss
```diff
@@ -76,6 +76,54 @@
   padding-top: u(.5rem);
 }
 
+.report-type-col {
```
noahmanger: More of a stylistic thing, but what about renaming to `.column---download` and putting it in `_datatables.scss` where the other `.column---*` classes live?

scss/modules/_data-container.scss
```diff
@@ -76,6 +76,54 @@
...
+    }
+
+    .dropdown {
+      position: initial;
+    }
+
+    .dropdown__panel {
+      right: u(1rem);
+      left: auto;
+      margin-top: u(.15rem);
```
xtine: Good catch. I wasn't seeing it because I was looking at it the large breakpoint. Using `top` property values instead for more accuracy.

noahmanger:
I'm still seeing the same problem on the dropdowns:
image

xtine:
Sometimes I notice this too when I'm changing sizes of the browser, but if you refresh within that size it resets the top spacing of the dropdown.

noahmanger:
Oh I see what's happening. It's positioned relateive to the cell, so when the text breaks onto two lines and makes the cell taller it changes:

demo

Is there a different way to do it?

noahmanger:
Since @xtine is out I went ahead and fixed these couple of things.

PR #551, opened 2016-10-19
Adding definition for other group

noahmanger:
Adds a glossary term for "Other groups" with this definition:

```
Form 5, Form 7 and Form 9 filers, whose activity includes: 
- Contributions reported by persons other than political committees
- Independent expenditures reported by persons other than political committees
- Communication costs reported by corporations and membership organizations
- Electioneering communications
```

cc @emileighoutlaw @amykort

### DIFF
--- a/js/terms.json
+++ b/js/terms.json
@@ -113,7 +113,7 @@
     "definition": "An organization, agency or employee representative committee or plan, in which employees participate and which exists for the purpose of dealing with employers on grievances, labor disputes, wages, hours of employment or working conditions. 114.1(d)."
   },
   {
-    "term": "Other groups",
+    "term": "Communications filers",
     "definition": "Form 5, Form 7 and Form 9 filers, whose activity includes: <ul class='list--bulleted'><li>Contributions reported by persons other than political committees</li><li>Independent expenditures reported by persons other than political committees</li><li>Communication costs reported by corporations and membership organizations</li><li>Electioneering communications</li></ul>"
   },
   {
--- a/js/terms.json
+++ b/js/terms.json
@@ -113,6 +113,10 @@
     "definition": "An organization, agency or employee representative committee or plan, in which employees participate and which exists for the purpose of dealing with employers on grievances, labor disputes, wages, hours of employment or working conditions. 114.1(d)."
   },
   {
+    "term": "Other groups",
+    "definition": "Form 5, Form 7 and Form 9 filers, whose activity includes: <ul class='list--bulleted'><li>Contributions reported by persons other than political committees</li><li>Independent expenditures reported by persons other than political committees</li><li>Communication costs reported by corporations and membership organizations</li><li>Electioneering communications</li></ul>"
+  },
+  {
     "term": "Person",
     "definition": "An individual, partnership, political committee, corporation, labor organization or any other organization or group of persons, not including the federal government. 100.10."
   },


jenniferthibault:
A thought for when we have these forms hosted on beta: would we want to link to those forms  (or wherever on the site they are introduced) where they appear in the definition here, so that people can look up what each form is/means?

Even after being on this project for a while now, I have a hard time remembering what each form number does without the full form title, so it might help novice users to have a way into looking that up.

emileighoutlaw:
I think that's a good thought, Jen— although I'd like to hold on this until we have forms actually on the beta site (it seems like it would be less helpful to have this functionality if it redirects users to the legacy FEC.gov). 

How about I add your note to this existing issue? https://github.com/18F/fec-style/issues/431

emileighoutlaw:
If that ^^ sounds good to you, Jen and if @amykort gives this a thumbs up to this PR, I'll merge!

jenniferthibault:
Sounds good, and totally agree about usefulness of forms on beta. 
Glad to add it to the existing issue, or we could make a smaller ice boxed issue if you wanted to close the existing one based on this PR. Whatever you prefer!

AmyKort:
This is to identify "other groups" for the purposes of the data visualization? If that's the case, I have a concern about making it a glossary entry. "Other groups" is not a legal term of art. It is, though, a pretty handy phrase. I don't feel confident that we aren't using "other groups" other places on the site where it might refer to a different bunch of people. @PaulClark2

emileighoutlaw:
Happy to hear Paul's thoughts, but I did do a look around the website, and we don't have "Other groups" anywhere else (yet).

We can also tweak the content so that it's clear we're talking only about "other groups" in data visualization.

AmyKort:
It does appear in the campaign guides, folks use the term in AORs that I found in a quick search of our materials, I think we'll find it in Record articles, etc. I'm hesitant to define it without making clear that the definition only applies to the data visualization.

emileighoutlaw:
Ah, that makes sense. Hmmmmm, that throws a wrench in the works. I'm going to think on this for a second.

noahmanger:
This sounds like it might be a case for a tooltip? Which would establish the rule that tooltips are for localized clarification of UI elements?

PaulClark2:
@AmyKort @emileighoutlaw how about "communications filers?"  The Press Office has received approval to use this term to describe Form 5, Form 7 and Form 9 filers collectively.  

http://www.fec.gov/press/summaries/2016/ElectionCycle/18m_IE_EC.shtml
http://www.fec.gov/press/press2016/news_releases/20160909release.shtml

AmyKort:
Paul is so awesome . . .

emileighoutlaw:
I keep feeling like I want to find a way to make a glossary definition work. What if it become very specific, like so:


>**Other groups (data visualizations)**
>In [Raising]() and [Spending]() charts, "other groups" are Form 5, Form 7 and Form 9 filers, whose activity includes: 
- Contributions reported by persons other than political committees
- Independent expenditures reported by persons other than political committees
- Communication costs reported by corporations and membership organizations
- Electioneering communications


This way the term name is more specific and the definition links to the specific places that one might find this term. I _believe_ that helps make clear that we aren't talking about anything on the website called "other groups," but I do want to hear what other folks think. 

cc @AmyKort

AmyKort:
I think Paul was suggesting we change up the data visualization to refer to those filers as "communications filers" and then define that term in a glossary definition. What do you think of that idea?

emileighoutlaw:
Ah! I somehow missed y'alls notes because my email is behind. YES! Paul is the best. 


We can call them "Communications filers" on the chart and then the glossary definition is:

```
**Communications filers**
Form 5, Form 7 and Form 9 filers, whose activity includes: 
- Contributions reported by persons other than political committees
- Independent expenditures reported by persons other than political committees
- Communication costs reported by corporations and membership organizations
- Electioneering communications
```

emileighoutlaw:
(Just committed that change)

noahmanger:
Looks good. Thanks everyone! (Merging this since it really only contains @emileighoutlaw 's commit).

PR #557, opened 2016-10-24
Eliminates view

emileighoutlaw:
"View" generally isn't an ideal word for interface copy, because it can feel exclusionary to folks who use screenreaders and aren't "viewing" the text. 

This changes "view" to "show"


ccostino:
Thank you, @emileighoutlaw! :-)

PR #561, opened 2016-10-28
Search consistency

emileighoutlaw:
Improves consistency of search boxes .Removes placeholder text per: https://github.com/18F/fec-cms/issues/531

Content only. Anyone can review.

### DIFF
--- a/js/templates/mobile-nav.hbs
+++ b/js/templates/mobile-nav.hbs
@@ -8,7 +8,7 @@
     <li class="site-nav__block">
       <a class="site-nav__link--underline" href="{{webAppUrl}}">Profiles</a>
       <form action="{{webAppUrl}}" method="GET" autocomplete="off">
-        <label class="t-sans" for="menu-search">Search for candidates and committees</label>
+        <label class="t-sans" for="menu-search">Search for candidates</label>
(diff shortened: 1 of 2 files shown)


js/templates/mobile-nav.hbs
```diff
@@ -11,7 +11,6 @@
         <label class="t-sans" for="menu-search">Search for candidates and committees</label>
         <div class="combo--search--mini">
           <input id="menu-search" name="search" class="combo__input js-menu-search" type="text"
-          placeholder="Search for candidates"
```
jenniferthibault: I know why we're doing this, but want to nudge that this may make a confusing interaction for folks since you cannot search across candidates AND committees at the same time yet: the search isn't integrated. In this case, the placeholder text reinforced you were searching just for candidates. Do you still want to go ahead @emileighoutlaw ?

emileighoutlaw: Oh so interesting! Actually, only the mobile nav says "Search for candidates and committees" the desktop nav says "Search for candidates" (and omits committees, which people can't search for now. I'm going to change the mobile navigation to match the desktop (i.e., leave off the "and committees" part)

PR #565, opened 2016-11-07
Press boilerplate

noahmanger:
This adds specific styles for the boilerplate paragraph at the end of press releases. For new press releases, this will only be visible when printing, when it will look like this:

image

For pre-2016 press releases, this paragraph isn't styled any differently and will remain in the web view, as it's just part of the content area.


ccostino:
Thanks, @noahmanger!

PR #566, opened 2016-11-08
Individual contributions styles

xtine:
Styles to support breakdown charts on 18F/openFEC-web-app#1665

### DIFF
--- a/scss/components/_breakdowns.scss
+++ b/scss/components/_breakdowns.scss
@@ -3,14 +3,6 @@
   padding: u(0 2rem);
   position: relative;
   width: 100%;
-
-  .sidebar-container {
-    max-width: 221px;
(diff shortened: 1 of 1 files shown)


scss/components/_breakdowns.scss
```diff
@@ -1,3 +1,18 @@
+.breakdown-container {
+  @include clearfix();
+  padding: u(0 2rem);
+  position: relative;
+  width: 100%;
+
+  .sidebar-container {
```
noahmanger: This and the `.main__content--right` selectors should be within a media query so that it stacks full width on mobile.

jenniferthibault:
Is this still relevant?

noahmanger:
Let's close this.

PR #574, opened 2016-11-28
Set width of filters__title

noahmanger:
Ensures that the filters icon, arrow and title all line up neatly, like:

image

https://github.com/18F/openFEC-web-app/pull/1711

cc @jenniferthibault @xtine

### DIFF
--- a/scss/modules/_filters.scss
+++ b/scss/modules/_filters.scss
@@ -34,6 +34,8 @@ $filter-button-width: u(4.6rem);
       left: 0;
     }
 
+    // For long filter titles, wrap them in a class to force icons to line up
+    // 7rem = width of filter icon + arrow icon + margin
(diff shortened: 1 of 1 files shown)


scss/modules/_filters.scss
```diff
@@ -34,6 +34,13 @@ $filter-button-width: u(4.6rem);
       left: 0;
     }
 
+    // For long filter titles, wrap them in a class to force icons to line up
+    // 7rem = width of filter icon + arrow icon + margin
+    .filters__title {
+      display: inline-block;
+      width: calc(100% - 7rem)
```
xtine: Just needs this fix and should be good to go.

noahmanger: Oops. Fixed!

PR #575, opened 2016-11-28
Adjusting size and positioning of inline icons

noahmanger:
This makes a couple necessary adjustment in order to display inline icons as in this instance:
image

1. The `.icon` class now adjusts the background image to be 100% of the element
2. The `.icon--inline--*` classes are `display: inline-block`

Goes with https://github.com/18F/openFEC-web-app/pull/1707

### DIFF
--- a/scss/components/_icons.scss
+++ b/scss/components/_icons.scss
@@ -22,11 +22,11 @@
 .icon--inline--right {
   display: inline-block;
   margin-left: u(1rem);
-  vertical-align: middle;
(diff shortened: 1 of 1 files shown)


xtine:
This looks good, although I wonder if the inline icon should also vertically align? Looks a bit off to me.

noahmanger:
Good eye. I just changed it to `vertical-align: text-bottom` which is more centered (border around the `<p>` for illustration):

image

PR #577, opened 2016-12-08
Fix calendar filters

noahmanger:
This fixes https://github.com/18F/fec-cms/issues/618 , an issue caused by the way we were handling multiple filters that share the same `name`. 

On the calendar page, every filter except for "State" filters the API by the `category` parameter. The way we handle this is to have the `name` of the form element set to the property it filters. So all of these checkboxes are like `<input type="checkbox" name="category" value="report-Q">`.

However! The filter code we've written for everything else assumes a _unique_ `name` value for each filter component. Without it, things like the incrementing/decrementing of the count, the filter tags, and loading from the URL don't work. 

We've gone around in circles with this particular headache a couple times now. I _think_ I finally have a good solution. 

This change reintroduces the `MultiFilter` component, but with a few changes:
- MultiFilters contain subfilters, which are instances of the `CheckboxFilter` class
- MultiFilters are instantiated by the `FilterSet` class, but subfilters are not instantiated until later. This is because the `FilterSet` can't handle multiple instances of a `Filter` object with the same `name`
- Instead, subfilters are then instantiated by the `MultiFilter` and are stored in an array as a property
- The `MultiFilter` also has its own `fromQuery` method that passes the query through to each `CheckboxFilter` subfilter
- Last, subfilters need to have their `filterLabel` (i.e. the accordion button that contains the number) explicitly defined on their element), so this gets set when they're instantiated

This feels a little over-engineered, but after working on it for the better part of today I'm just not sure what a better approach is, but certainly open to alternatives.

### DIFF
--- a/js/checkbox-filter.js
+++ b/js/checkbox-filter.js
@@ -43,6 +43,7 @@ CheckboxFilter.prototype.handleChange = function(e) {
       key: id,
       value: this.formatValue($input, $label.text()),
       loadedOnce: loadedOnce,
+      filterLabel: this.$filterLabel,
       name: this.name
     }
   ]);
--- a/js/filter-base.js
+++ b/js/filter-base.js
@@ -29,10 +29,6 @@ function Filter(elm) {
   this.fields = [this.name];
   this.lastAction;
 
-  $(document.body).on('filter:added', this.handleAddEvent.bind(this));
-  $(document.body).on('filter:removed', this.handleRemoveEvent.bind(this));
-  $(document.body).on('filter:changed', this.setLastAction.bind(this));
-
   if (this.$elm.hasClass('js-filter-control')) {
     new FilterControl(this.$elm);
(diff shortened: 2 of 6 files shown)


noahmanger:
Ugh I am completely stumped as to why the tests are failing. Things work as expected in the wild. I've traced the issue down to [this line](https://github.com/18F/fec-style/pull/577/files#diff-b295643d945c45d2c6bc8022e59e1087R82) but I have no idea what the problem is. Any ideas @xtine?

noahmanger:
Nevermind! Figured out the tests. Those pesky event listeners had to be removed on `beforeEach`. :shakes-fist:

xtine:
Looks good! I like the use of `filterLabel`. I think this is a suitable solution given that the hack I did with trying to identify filters other than with `name` wasn't going to fully work.

PR #579, opened 2016-12-12
New homepage header + hero styles

xtine:
In support of 18F/fec-cms#655


scss/modules/_hero.scss
```diff
@@ -43,6 +43,58 @@
...
+
+  .hero__banner {
+    background-image: url('../img/fec-table-tint.png');
+    width: 100%;
+    height: u(22rem);
+    position: absolute;
+  }
+
+  h2 {
+    color: $inverse-translucent;
```
noahmanger: I _think_ this should actually just be white, so `$inverse`. I think the shading you see is an artifact of the jpeg maybe? @jenniferthibault can you verify that the main title over the image of the chairs should be full opacity white?

scss/modules/_site-header.scss
```diff
@@ -72,6 +72,14 @@
   }
 }
 
+.site-header-homepage .disclaimer {
```
noahmanger: For BEM consistency, can this be `.site-header--homepage` (double --).

jenniferthibault:
I can't find the comment thread in this issue, but in a GitHub notification @noahmanger asked me: 
`I think this [text] should actually just be white, so $inverse. I think the shading you see is an artifact of the jpeg maybe? @jenniferthibault can you verify that the main title over the image of the chairs should be full opacity white?`

I know this has already been merged, but I can't tell if that was settled or not. But yes, there shouldn't be any transparency on text, and the image is already edited to create the contrast balance needed between text and background. There shouldn't be any image or text adjustments needed in the code.

noahmanger:
Cool. @jenniferthibault, @xtine made the change before I merged.

jenniferthibault:
Great! Yeah that's what it looked like, I just wanted to do my part to confirm :)
Thanks you two!

PR #582, opened 2016-12-14
[WIP] Search.js and updated homepage styles

patphongs:
## Summary

Search.js was moved into the fec-style repo so that it may be used globally. Also added a homepage.scss file for specific homepage styles. We may want to revisit this later if we find that these styles will be used in places other than homepage.


xtine:
I would suggest to set left/right padding on homepage sections so it lines up with the new homepage specific header, like so:
https://github.com/18F/fec-style/blob/master/scss/modules/_hero.scss#L93

scss/modules/_homepage.scss
```diff
@@ -3,15 +3,61 @@
 // Campaign Finance Data Homepage Styles
 .slab.slab--primary .blue {
```
xtine: I would suggest to add add a `slab--homepage` class, so then this could be defined as `.slab--primary.slab--homepage`, so then the class can be re-used for specific homepage slab coloring beacuse the link color also changes for the secondary slab on the homepage.

scss/layout/_grid.scss
```diff
@@ -138,16 +149,19 @@
     }
 
     .grid--2-wide {
-      .grid__item:nth-last-child(-n+2) {
-        border-bottom: none;
-        margin-bottom: 0;
+      .grid__item:nth-child(1),
```
xtine: I just realized that we could use the `nth-child(-n+2)` rule to grab the first two child elements, and ``nth-child(-n+3)` for three, etc.

scss/layout/_grid.scss
```diff
@@ -68,6 +69,16 @@
   }
 }
 
+.slab--primary .grid {
+  a {
```
noahmanger: I think we should consider putting either these link styles in their own class or making a class for these slabs on the home page. Elsewhere, on a dark background, we use white links. This would make it so that any time a link is in a `.grid` container within a `.slab` the links would be a color. 

@jenniferthibault do you have thoughts on what the rules ought to be for when a link is white vs a contrasting color?

scss/layout/_grid.scss
patphongs: I updated the code to reflect this. thanks for the tip.

scss/layout/_grid.scss
jenniferthibault: Can you point me to the other pages that use `.grid` inside a `.slab` that you're seeing at issue, @noahmanger ? It could be easy and logical to eliminate any special instances, but these opportunities are only things I learn about through implementation like this. Glad to give it a go if we can surface the instances it would interfere with.

xtine: @jenniferthibault: the bottom of the press page uses cards in a grid in a slab.
screen shot 2016-12-20 at 3 51 31 pm

I'm in favor of merging this in and then making a special border class for grids instead of automatically having grids bordered in slabs in a later branch (with the other homepage design tweaks) to get all the homepage grids updated.

noahmanger:
I think this last commit erased my earlier comment. The thing I wanted to point out here is actually that this implements a new design for the search bar / select pattern than we use other blue backgrounds. You're implementing it the way it was designed, but I just wanted to make sure that was intentional on @jenniferthibault 's part. 

Do we want it to be this way (as Pat implemented and Jen designed):
image

Or this way, which is how it currently looks, using the default styling of selects and inputs on [navy backgrounds](https://beta.fec.gov/data/candidate/P00003392/):
image

scss/layout/_grid.scss
noahmanger: @xtine I'm fine with that having that change in a follow up, but we just need to not lose track of it.

noahmanger: Ships crossing in the night! I just commented about some of that, but a separate thought is: if we do need separate styles for the home page slabs, which it seems like we do for the borders and link colors, maybe we should just add `.slab--home` as a modifier class that could be added to `.slab--primary` or `.slab--secondary`.

jenniferthibault:
This comment thread got pretty huge and gnarly, so I hope I'm responding to the complete list of things you need. If not, please let me know. 

## Search bar
Please use the existing default color scheme for the default styling of selects and inputs on [navy backgrounds]. This was my mistake, I am losing track of patterns. (https://beta.fec.gov/data/candidate/P00003392/):
image

## Link color in slabs
@noahmanger asked:
> What is the rule for when a link on the primary or secondary background is the contrasting color vs. when it's white? By default, these links are white, for example, in breadcrumbs or committee headers. But here we're making them contrast colors.

As I see it, it's the rule when the link is the primary call to action.

scss/elements/_forms.scss
```diff
@@ -240,3 +240,48 @@ button {
     opacity: 0.5;
   }
 }
+
+// Homepage candidate/commitee search box styles
+
+.slab--primary .grid__item .js-search-type {
```
patphongs: Thanks for clarifying @jenniferthibault. I removed the additional styles and the search section is now defaulting to the original styles.

scss/layout/_grid.scss
patphongs: thanks for the feedback. I've added the additional .slab--home modifier class

noahmanger:
Excellent!

PR #586, opened 2016-12-15
New Homepage Commissioners section styles

xtine:
Support for https://github.com/18F/fec-cms/pull/662


scss/layout/_grid.scss
```diff
@@ -119,14 +130,32 @@
...
   }
 
-  .slab .grid {
-    .grid__item:nth-last-child(-n+2) {
-      border-bottom: none;
-      margin-bottom: 0;
+  .slab {
+    .grid {
+      h3 {
+        line-height: 1.5;
```
noahmanger: Feels a little weird to have this property nested in `.slab > .grid`. Should it be a more generalized class or applied to all h3s?

scss/layout/_grid.scss
```diff
@@ -119,14 +130,32 @@
...
+      }
+    }
+
+    .grid--2-wide {
+      .grid__item:nth-last-child(-n+2) {
```
noahmanger: I didn't know about the `-n` trick.

scss/layout/_grid.scss
```diff
@@ -119,14 +130,32 @@
     }
   }
 
-  .slab .grid {
-    .grid__item:nth-last-child(-n+2) {
-      border-bottom: none;
-      margin-bottom: 0;
+  .slab {
```
noahmanger: Can this be combined with the `.slab > .grid` nest above?

scss/layout/_grid.scss
xtine: Yeah I'm not quite sure where to put this because I don't want to affect the other existing `h3`.

ccostino:
Since this is approved and remarked, I'm merging in - we can continue to improve in the release branch if need be!

PR #589, opened 2016-12-19
Legal resources icons for new homepage

johnnyporkchops:
Added icon styles for legal resources section on new homepage


noahmanger:
This looks great. I cleaned up some of the errors that Hound was barking about.

PR #593, opened 2016-12-22
Data type toggle

noahmanger:
This does a couple primary things to support the new efiling UI.

1. In `filter-panel.js` it distinguishes between the `toggle` (which is always the top button, and which is used to open and close the panel) and the `filterHeader`, which is the heading for the filters that contains the number of applied filters.

2. In `filter-set.js` I refactored the activation code so that when a filter-set has an efiling set of filters (determined by having `data-efiling-filters` set on the element), it activates the processed and the efiling and processed filters _separetly_ and stores them in distinct properties. 

That way, when `filterSet.activateSwitchedFilters()` is called, it can only set the values of those filters in the visible set. This method also stores the previous state (from the query) in order to repopulate the filters when toggling back.

3. `filter-set.js` is also the place where the visibility of the filters is toggled. `filterSet.switchFilters()` is called in `tables.js` over in `openfec-web-app`. 

4. Last, to clear all filters, `filterSet` just triggers the `tag:removeAll` event. 

Oh, and I also updated the PhantomJS number because my MacOS Sierra wasn't working with it, and this seemed to be the solution.

### DIFF
--- a/js/dropdowns.js
+++ b/js/dropdowns.js
@@ -157,10 +157,11 @@ Dropdown.prototype.handleRemoveClick = function(e, opts) {
 // "Clear all filters" will remove unchecked dropdown checkboxes
 Dropdown.prototype.handleClearFilters = function() {
   var self = this;
-
-  this.$selected.find('input:checkbox:not(:checked)').each(function () {
-    self.handleCheckboxRemoval($(this));
-  });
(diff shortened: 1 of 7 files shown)


noahmanger:
Also ready for review.

PR #596, opened 2016-12-28
Added About hero image

patphongs:
Hero image for the about page has been added per issue #667


package.json
```diff
@@ -23,10 +23,11 @@
     "test-single": "karma start --single-run --browsers PhantomJS"
   },
   "dependencies": {
+    "@frctl/fractal": "^1.0.11",
```
noahmanger: This looks like an accidental inclusion. We don't need to include fractal in fec-style since it's its own repo.

patphongs: You're right, thanks, have removed it.

PR #598, opened 2016-12-31
Home grid tweak

noahmanger:
Makes a few small things to account for things in the citizens info slab on the home page:

image


noahmanger:
Goes with https://github.com/18F/fec-cms/pull/697

noahmanger:
Got it!

PR #605, opened 2017-01-11
Complete, larger images for all commissioners

jenniferthibault:
## Summary

For https://github.com/18F/fec-cms/issues/702

Adds images of commissioners at large size (136x136) for commissioner bio page templates.

These images aren't great quality, but they'll be ok to start. @patphongs is tracking down the originals, which will give us better quality and I'll later replace these.

@xtine this will replace the smaller images of the six current commissioners, and may cause sizing issues, but guessing you're expecting that based on https://github.com/18F/fec-cms/issues/702#issuecomment-271027541

## Screenshots

screen shot 2017-01-11 at 1 00 52 am


noahmanger:
Rad. Because these are just going to be used in the CMS, I think we can just add them there. What do you think @xtine ?

noahmanger:
Actually, scratch that. These will be uploaded as images in the CMS, so we don't need to check them in anywhere.

xtine:
As @noahmanger mentioned, these don't need to be added to the repo since they are going in the CMS. However, I did upload into my local CMS the images of former Commissioners Harris and Mason to test how the bigger images show up. They look great!

screen shot 2017-01-11 at 3 33 32 pm
screen shot 2017-01-11 at 3 33 26 pm

PR #607, opened 2017-01-12
Filter tag update

noahmanger:
This change is to accommodate the fact that when switching between processed and raw filters, sometimes tags that are otherwise non-removable (like dates) will need to be removed because the filter doesn't exist on the other data type. For example, this is the processed view for receipts:
image

And this is raw:
image

To do this I added:
- A new option to pass in the `tag:removeAll` event for `forceRemove` which tells the filter tags to remove any that can be removed on the data type switch
- I added a new data attribute `data-remove-on-switch` that's added to "nonremovable" tags, which means that they can be cleared when `forceRemove = true`. 
- Added a remove handler to the date filter to actually clear those filters

### DIFF
--- a/js/date-filter.js
+++ b/js/date-filter.js
@@ -82,7 +82,7 @@ DateFilter.prototype.handleInputChange = function(e) {
       rangeName: rangename,
       name: this.name,
       nonremovable: nonremovable,
-      removeOnSwitch: true,
+      removeOnSwitch: true
     }
   ]);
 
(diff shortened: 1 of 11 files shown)


js/date-filter.js
```diff
@@ -80,7 +81,8 @@ DateFilter.prototype.handleInputChange = function(e) {
       range: range,
       rangeName: rangename,
       name: this.name,
-      nonremovable: nonremovable
+      nonremovable: nonremovable,
+      removeOnSwitch: true,
```
xtine: trailing comma

PR #609, opened 2017-01-13
Run styleguide on cloud.gov

adborden:
Pages is going away, moving to a static buildpack on cloud.gov.
https://fec-style.apps.cloud.gov/


noahmanger:
Apologies for any confusion that led to duplicate work, but since you got the pattern library set up (https://github.com/18F/fec-pattern-library/pull/3) we don't need this anymore, right?. Should this just be updated to remove the 18F pages step?

adborden:
@noahmanger no worries, it wasn't much work at all. I wasn't sure if everything from fec-style had already been migrated.

noahmanger:
Cool. Yep it's all covered. Now we get to clean out the old KSS code. Thanks again!

PR #611, opened 2017-01-14
Fixing the accuracy of the filter count

noahmanger:
This fixes a few unrelated bugs that occur when toggling between raw and processed filters which cause the number in the filter panel to be inaccurate:

- The `FilterSet` no longer activates both sets of filters on initalization. Instead, it activates the `DataType` toggle, which then activate whichever set of filters it needs. This is because some filters have default values when initialized, so they were being initialized -> triggering `filter:added` events -> inflating the number, but weren't actually visible
- When clearing date filters, it now triggers the `filter:removed` event to decrement the number
- The `ToggleFilter` class now can have a `ignoreCount` property set, and if this is passed in the `filter:added` event to `FilterSet` then the count isn't incremented. This is because we don't want the DataType filter to actually increase the number. This requires a small change on the web app.

### DIFF
--- a/js/date-filter.js
+++ b/js/date-filter.js
@@ -152,11 +152,16 @@ DateFilter.prototype.handleRemoveAll = function(e, opts) {
   // If this is a forceRemove event that means it was triggered by table switch
   // So we need to clear these inputs and set had-value to false so that it fires filter:added
   var forceRemove = opts.forceRemove || false;
+
+  function remove($filter) {
+    $filter.val('');
+    $filter.data('had-value', false);
+    $filter.trigger('filter:removed', {loadedOnce: true});
+  }
+
   if (forceRemove) {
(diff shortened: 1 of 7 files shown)


js/date-filter.js
```diff
@@ -152,11 +152,16 @@ DateFilter.prototype.handleRemoveAll = function(e, opts) {
...
+    $filter.data('had-value', false);
+    $filter.trigger('filter:removed', {loadedOnce: true});
+  }
+
   if (forceRemove) {
-   this.$minDate.val('');
-   this.$maxDate.val('');
-   this.$minDate.data('had-value', false);
-   this.$maxDate.data('had-value', false);
+   remove(this.$minDate);
```
xtine: Nice, I like this creation of the remove function.

Issue #616, opened 2017-01-19
Make icon sizes consistent on AO Process page


jenniferthibault:
https://beta.fec.gov/legal/advisory-opinions/process

The email icon is smaller than the others: the hand delivery & fax icons are the correct size.

<img width="569" alt="screen shot 2017-01-19 at 9 44 09 am" src="https://cloud.githubusercontent.com/assets/11636908/22111064/f99d09e0-de2b-11e6-8d23-bcefead2e831.png">

Not sure what is causing this, but if it's something I can help with let me know!

cc @xtine

AmyKort:
Wholly unrelated: can you take Shawn's name off of the hand delivery address and have it just go to the office of the commission secretary? Thanks!

jenniferthibault:
🎉 Removing the name has the side effect of making the text columns more even in length. Definite visual perk!

I also just noticed that the email address next to the email icon looks like a smaller font size. Weird things happening with that one! 

(Have I mentioned lately that it's awesome having you on GitHub Amy?)

AmyKort:
Aw shucks. Thanks @jenniferthibault !

noahmanger:
Removed Shawn's name and put a fix for the icon sizing in at #617

PR #617, opened 2017-01-20
Fixing icon size

noahmanger:
Fixes the sizes of the contact-item icons to be consistent:

image

Resolves https://github.com/18F/fec-style/issues/616

### DIFF
--- a/scss/components/_contact-items.scss
+++ b/scss/components/_contact-items.scss
@@ -7,7 +7,7 @@
     &::before {
       content: '';
       display: block;
-      background-size: 90%;
+      background-size: 100%;
       margin-right: 2rem;
       width: u(3.4rem);
       height: u(3.4rem);
@@ -77,8 +77,8 @@
 .contact-item--hand {
   &.contact-item::before {
     @include u-icon-circle($hand-envelope, $primary, $inverse, 3.4rem);
-    background-size: 100%;
     background-position: 100% 50%;
+    background-size: 100%;
   }
 }
 


jenniferthibault:
The things:

- Is there a reason that the items got re-ordered? I may have missed this in discussion somewhere. 
- The text size of the content under `Hand delivery` looks smaller than the rest, and it shouldn't be

noahmanger:
Oops, fixed the sizing. The ordering is because of what I showed you a while ago where because these are components that the template is generating, they flow in a different order. Keeping them in the same order would result in this (because they order horizontally rather than in columns):

image

What would you prefer?

jenniferthibault:
The order does indicate some level of preference to me, so I'd like to double-check with @emileighoutlaw that the preferred method is to send by email, rather than hand delivery.

I see what you mean about the spacing, but I think our grid keeps the items well organized here, and it doesn't feel chaotic to me. Rather, if there were a list of 20 items, I'd prefer them to be aligned in pairs horizontally, like this hints at.

emileighoutlaw:
This is a great question that I actually don't know the answer to. Checking now.

emileighoutlaw:
Confirmed with the FEC that preferred method is by email!

noahmanger:
This PR only handles the size of the icons. The order is set in the CMS. So this can be merged if the sizing looks right.

jenniferthibault:
Sorry @noahmanger in your most recent screenshot the sizing looks incorrect so I'm not sure where this stands (if that was an old screenshot or not). Let's pair on this real quick today?

noahmanger:
Oops, that was just an old screenshot. Here's how it would actually look (both 34px):
image

jenniferthibault:
🙌  thank you! merging

PR #621, opened 2017-02-01
Fix filter quotes

noahmanger:
When text and typeahead filters are generate checkboxes, they previously weren't stripping quotes out of the values, which caused the `<input type="checkbox">` to have an empty `value` attribute. This fixes that, refactors the data formatting slightly, and adds a couple tests.

I have one thing left to do on this, but need to step away and wanted to submit before I forgot.

### DIFF
--- a/js/filter-typeahead.js
+++ b/js/filter-typeahead.js
@@ -19,8 +19,17 @@ function slugify(value) {
 function formatLabel(datum) {
   return datum.name ?
     datum.name + ' (' + datum.id + ')' :
-    '"' + datum.id + '"';
+    '"' + stripQuotes(datum.id) + '"';
 }
+
+function formatId(name, value) {
+  return name + '-' + slugify(value) + '-checkbox';
+}
+
+function stripQuotes(value) {
+  return value.replace(/["]+/g, '');
+}
+
 
 var textDataset = {
   display: 'id',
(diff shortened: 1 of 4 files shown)


xtine:
Should we also strip out `'`?

noahmanger:
I thought about it, but I don't think so. There may be apostrophe's in employer names or individual names, and they won't break the `<input value>`.

xtine:
What about checking if it's just single quotes at both beginning and end? So it would strip `'Portland'` but not `Portland's`.

noahmanger:
@xtine how easy is that to do? My regex chops are...not good :) . And do you think it's important to do that?

xtine:
@noahmanger: it's probably more efficient to just check if first/last characters are `'` and remove them than to scan through the entire string using a regex. But I think it might be worth it if users are already searching for `"Portland"` and getting 0 results, might as well make sure `'Portland'` isn't affected as well.

noahmanger:
I don't think that's actually necessary. Including quotes isn't the issue... the API ignores them and returns results regardless. The issue was that double quotes were resulting in checkboxes like `<input type="checkbox" value=""portland"">` (where `value=""`)

xtine:
@noahmanger: gotcha.

PR #625, opened 2017-02-07
Fix firefox sidenav

noahmanger:
This fixes a long-standing bug with the sticky sidenav in Firefox where clicking the links wouldn't work. Apparently Firefox can't animate the `body` element, only `html`. 

demo

Second, it adjusts the scroll setting so that the correct item is highlighted at the correct point.

Third, I also added an antialiasing property to `body` so that the font-rendering is consistent with Chrome. Previously, light text on dark background looked extra bold.

Before:
image

After:
image

Resolves https://github.com/18F/fec-cms/issues/280
Resolves https://github.com/18F/fec-cms/issues/244
Resolves https://github.com/18F/FEC/issues/575
Resolves https://github.com/18F/FEC/issues/586

### DIFF
--- a/js/toc.js
+++ b/js/toc.js
@@ -56,7 +56,7 @@ TOC.prototype.scrollTo = function(e) {
   e.preventDefault();
   var $link = $(e.target);
   var section = $link.attr('href');
-  var sectionTop = $(section).offset().top;
+  var sectionTop = $(section).offset().top + 10;
   $('body, html').animate({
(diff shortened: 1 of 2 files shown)


johnnyporkchops:
This problem was also happening in IE11, I will test to see if this solution may have fixed it there as well.

jenniferthibault:
I'm very excited about this!

PR #627, opened 2017-02-08
Contact form styles

noahmanger:
Adds some styles for the contact form in https://github.com/18F/fec-cms/pull/755#issuecomment-278198318

### DIFF
--- a/scss/components/_contact-form.scss
+++ b/scss/components/_contact-form.scss
@@ -9,12 +9,6 @@
     height: u(15rem);
   }
 
-  .conditional-field {
-    border-left: 8px solid $gray;
-    padding: u(2rem);
-    margin-bottom: u(2rem);
-  }
-
   .tt-input {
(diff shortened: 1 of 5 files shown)


scss/components/_contact-form.scss
```diff
@@ -0,0 +1,52 @@
...
+  textarea {
+    height: u(15rem);
+  }
+
+  .tt-input {
+    @include u-icon-bg($magnifying-glass-arrow, $gray-dark);
+    background-position: right 10px top 50%;
+  }
+
+  .hr--soft {
```
xtine: Just kind of a minor nitpick, but I think the hr naming convention might make more sense to follow the color variables, so like `.hr` and `.hr--dark` (and `.hr--light` if using `$gray-light`). What do you think?

noahmanger: Good call.

PR #631, opened 2017-02-11
New committee pages

noahmanger:
This adds a few new styles for the new candidate and committee pages. 

First, there's styles for the financial summary component. To get this effect, a `level--*` class is applied to the `<tr>`:
image

It also includes styles for this pattern:
image

And some new tweaked styles for datatables on these pages:
image

### DIFF
--- a/scss/components/_figures.scss
+++ b/scss/components/_figures.scss
@@ -39,3 +39,21 @@
     display: inline !important;
   }
 }
+
+.entity__figure {
+  background-color: $inverse;
+  padding: u(1rem);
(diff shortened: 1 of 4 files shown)


xtine:
For the new version info on the datatable, can we add a space on the second line so the text aligns?

noahmanger:
@xtine got it! I replaced the `&nbsp;` with an empty icon class.

xtine:
Oooh blank icon class is brilliant. I'll add that onto filings too.

noahmanger:
I believe it's the same function, so already there!

Issue #633, opened 2017-02-15
[ATO] Address cross-site scripting and content-spoofing issues


ccostino:
The folks performing testing for our ATO have found several instances where it could be possible to initiate a cross-site scripting attack by adding JavaScript as values to the parameters in our web requests, or supplying additional content as a part of the URL request.

We need to make sure we're sanitizing user input appropriately and escaping things that need escaping to address this issue.

NOTE: This is intended for GovCloud, but needs to be fixed across the board.  This is also related to this issue: https://github.com/18F/openFEC-web-app/issues/1839

ccostino:
@xtine and I have identified [this line](https://github.com/18F/fec-style/blob/master/js/filter-set.js#L51) as the apparent root of the issue that we can fix.

ccostino:
I've done a bit more research and that line is just the root for the issue reported - it goes a bit deeper than that.  We need to sanitize the query string wherever it is parsed by URI.js.  Working on a solution that I'll put up as a PR soon!

noahmanger:
Resolved by https://github.com/18F/fec-style/pull/634

PR #634, opened 2017-02-16
Mitigate XSS and content spoofing attacks

ccostino:
Addresses #633

This changeset adds support to mitigate potential XSS and content spoofing attacks through our client-side code.  Penetration testing has yielded a few spots in our code where we were accepting values as is when pulling them from the query string in the URL, mainly when pre-initializing filters.  We have addressed this by making use of the `bleach` (https://github.com/ecto/bleach) library to strip all tags from this content, and then stripping out all other characters not whitelisted as being valid.  We determined that only certain characters should be allowed for the filter values and feel safe in ignoring all others.

A huge thank you to @xtine for pairing with me on this and helping solve the issue!  h/t to @LinuxBozo, @toolness, and @cmc333333 for the additional help!

### DIFF
--- a/js/analytics.js
+++ b/js/analytics.js
@@ -2,8 +2,10 @@
 
 /* global ga */
 
-var URI = require('urijs');
 var _ = require('underscore');
+var URI = require('urijs');
+
+var helpers = require('./helpers');
 
 function trackerExists() {
   if (typeof ga !== 'undefined') {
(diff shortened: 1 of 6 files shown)


ccostino:
I'm in the process of creating tests for these changes.  Once they're done this will be ready for review!

PR #635, opened 2017-02-21
Adding style for tiles

noahmanger:
This adds a new component which I called `.tile`, as used on the "Reports about the FEC" page. 

tile

The whole thing is a link. I wasn't sure if that was the behavior @jenniferthibault was expecting, but I like the nice big targets. I took a guess at what the visual styling for hovering should be, but that can change.


scss/components/_tiles.scss
```diff
@@ -0,0 +1,23 @@
+// Grid items that are also links
+
+.tile {
+  background-color: $gray-lightest;
+  border-top: 1rem solid $primary;
+  border-bottom: none;
+  display: block;
+  padding: 2rem 1rem;
```
jenniferthibault: Can we increase the left/right padding to 1.5rem please? Feels a little tight on the sides.

scss/components/_tiles.scss
```diff
@@ -0,0 +1,23 @@
...
+    min-height: 20rem;
+  }
+
+  @include media($lg) {
+    min-height: 25rem;
+  }
+
+  &:hover,
+  &:focus {
+    border-top-color: $primary-contrast;
```
jenniferthibault: When the panel is hovered over/focused on, could we also add dotted underline to the heading on that panel? That would reinforce our link pattern seen elsewhere, and would avoid relying on just a color change to an element.

noahmanger:
@jenniferthibault cool cool. I renamed to `card--major`.

noahmanger:
Also added an example to the pattern library https://github.com/18F/fec-pattern-library/pull/4

jenniferthibault:
@xtine could you take a look for technical review? Design looks good!

scss/components/_cards.scss
```diff
@@ -310,6 +310,52 @@
...
+// Style used when the cards are the primary feature of a layout,
+// such as the "Reports about the FEC" landing page.
+.card--major {
+  background-color: $gray-lightest;
+  border-top: 1rem solid $primary;
+  border-bottom: none;
+  display: block;
+  padding: 2rem 1.5rem;
+  position: relative;
+  transition: border-top-color .2s ease-out;
```
xtine: does this need `@include transition`?

PR #639, opened 2017-03-01
Update button

noahmanger:
Adds styles for these buttons:
image

I also renamed the previous `.button--calendar` to `.button--add-calendar`  to be descriptive of what it actually does:

image


jenniferthibault:
This calendar icon is really terrible at small sizes. We knew this before, but at this even spammer size it's really problematic. If you want to hold reviewing this PR, I can make one and add it in here tomorrow! I wouldn't call it critical though, so if  you'd rather I can make a follow up PR and this can keep going

noahmanger:
Please do! No rush for this. I just wanted to get it in before I forgot.

PR #642, opened 2017-03-07
Site nav update

noahmanger:
This implements the changes to the site nav as designed in https://github.com/18F/fec-cms/issues/827

- Adds mega menus for the other two sections
- Cleans up the campaign finance data menu
- Removes unnecessary styles that we don't need anymore with the simpler layout
- Adds new styles for the hover effects
- Adds an option that's used when initializing the mega menu to open it on click  

demo

Dependencies:
- Because the menu is an external library, [this PR](https://github.com/adobe-accessibility/Accessible-Mega-Menu/pull/45) needs to be merged and then an update pushed to NPM 
- This change for the CMS: https://github.com/18F/fec-cms/pull/854
- A similar change on the web app


jenniferthibault:
Thanks for getting this going so fast @noahmanger ! I used your screencapture for a visual review and have some tweaks. Glad to pair through them!

1. `Registration and reporting` should be `Candidate and committee services`
2. `Candidate and committee services` >> Information for: typo in`Corproations` » `Corporations`
3. In Legal resources, Policy should not be on the menu, since it would be an external link. See https://github.com/18F/fec-cms/issues/827#issuecomment-284492590
4. Order of legal resources items should be:
   - Advisory opinions
   - Enforcement
   - Statutes
   - Legislation
   - Regulations
   - Court cases

5. Section hover underline should be the width of the word, not the width of the container:

_Like this_
- mega-menu_hover-1
 
- screen shot 2017-03-07 at 9 43 46 am


_Instead of this_
- screen shot 2017-03-07 at 9 45 14 am
- screen shot 2017-03-07 at 9 45 05 am

6.  Categories should be `font-color: $base` (#212121 black, instead of federal blue)
7. Drop-down arrows should be `$grey-dark` (#5B616B) so that they are slightly lower contrast, and don't stand out so much. This will help balance the solid shape.
8. Like on the [USWDS](https://standards.usa.gov/components/headers/basic-mega/), I was thinking we should keep the menu item grid to the width of the main content container. The menu background should be full-screen-width, but the container should be narrower, which supports more predictable scanning, and reduces the space the mouse has to travel to navigate on large screens:
- screen shot 2017-03-07 at 10 09 41 am
- screen shot 2017-03-07 at 10 10 54 am

9. We should increase leading in list items so that when they underline the lines don't crash into the next text line (as seen in Reg/reporting) Can we pair quickly on this one? I think it will be a delicate scootch so that we don't accidentally increase the height of the container.

noahmanger:
Thanks @jenniferthibault . I'll take a pass and then loop you in for fine tuning.

jenniferthibault:
Carrying over copyedits from @emileighoutlaw in https://github.com/18F/fec-cms/issues/827#issuecomment-284909968

> Copyedit suggestion for the Candidate and committee services menu:
> - We'll want it to say _Information about_ rather than _Information for_.
> - I'd also suggest omitting the colon. I understand why it's there (and it's technically correct), but the fact that it's the only across the menus feels a little jarring. I don't believe omitting it will be a barrier to understanding and would make for a more consistent experience.

noahmanger:
Alright, I've made those edits. The container is still a tricky situation though, so I haven't changed that. It looks really odd to have the inner menu be in a fixed-width container but not the header. In the USWDS example, the header is also in a fixed-width container. So we should think more about this.

noahmanger:
Alright, after pairing with Jen, this is updated and ready for full review. The three sections:

image
image

image

Also, with the new simpler layout, it allows us to get rid of the separate templates for mobile and instead just use responsive styling to achieve a simpler layout like this:

image
image

emileighoutlaw:
This gets a content 👍 from me

scss/components/_mega-menu.scss
```diff
@@ -60,47 +120,31 @@
...
-    a {
-      border-bottom: 1px dotted $inverse;
-    }
   }
 
-  .mega__group {
-    @include clearfix();
-    padding-bottom: u(2rem);
+  .mega__page-link {
+    padding-left: 2rem;
```
xtine: u()

Issue #632, opened 2017-02-15
Design and implement block quote styles


jenniferthibault:
So that quotations can be pulled out from body copy in a way that makes them visually distinct from the rest of the body copy, design a style for them and implement it on the front end.

**Steps to completion**
- [x] Visual design styled: @jenniferthibault
- [x] Implemented in stylesheet
- [x] Added to Wagtail WYSIWYG editor for paragraph blocks

jenniferthibault:
What I was thinking for visual styling is super related to the Example styles pulled from https://github.com/18F/fec-cms/issues/727#issuecomment-276272557 . The type style would be standard body copy size, but italic weight:

![screen shot 2017-02-15 at 3 18 42 pm](https://cloud.githubusercontent.com/assets/11636908/22993728/11159ba2-f392-11e6-9525-395fac05a9ce.png)

I was also looking specifically at spacing because in the E-Filing report there are a few back-to-back block quotes, so this style would need to work in sequence:

![screen shot 2017-02-15 at 3 20 12 pm](https://cloud.githubusercontent.com/assets/11636908/22993813/6443ee5a-f392-11e6-9f3c-cced96332346.png)

noahmanger:
Resolved by https://github.com/18F/fec-cms/pull/861

PR #645, opened 2017-03-09
Added styles for block quotes

patphongs:
Related to https://github.com/18F/fec-cms/pull/861

Implementation of issue #632

### DIFF
--- a/scss/elements/_typography.scss
+++ b/scss/elements/_typography.scss
@@ -84,4 +84,6 @@ blockquote {
   border-left: 10px solid $gray-lightest;
   padding-left: 30px;
   font-size: u(1.6rem);
+  font-style: italic;
+  margin-left: 0;
 }
--- a/scss/elements/_typography.scss
+++ b/scss/elements/_typography.scss
@@ -79,3 +79,9 @@ hr {
   border-top: none;
   margin: u(1rem 0);
 }
+
+blockquote {
+  border-left: 10px solid $gray-lightest;
+  padding-left: 30px;
+  font-size: u(1.6rem);
+}


jenniferthibault:
So exciting! I 😻  new style work.

A couple minor adjustments based on the [initial direction](https://cloud.githubusercontent.com/assets/11636908/22993813/6443ee5a-f392-11e6-9f3c-cced96332346.png) :
- Font style should be _italic_
- No need for the left padding between the grey bar and the container—the bar should off-set it more than enough that we shouldn't need to give it extra space. I did a quick sketch of what I mean based on your screenshot:
screen shot 2017-03-09 at 9 25 07 am

scss/elements/_typography.scss
```diff
@@ -79,3 +79,9 @@ hr {
   border-top: none;
   margin: u(1rem 0);
 }
+
+blockquote {
+  border-left: 10px solid $gray-lightest;
+  padding-left: 30px;
+  font-size: u(1.6rem);
```
jenniferthibault: This is actually something I just don't know, but maybe you can teach me :) 
If this is inheriting the `<p>` class styles, do we need to set the font size? Or should it have it automatically?

patphongs: So the blockquote button is actually based inside of the paragraph wagtail component, however, the p tag does not actually wrap the blockquote so it does not inherit the paragraph styles. Which reminds me, I forgot to put in an image of what the blockquote button looks like in wagtail, see below.

screen shot 2017-03-09 at 9 50 21 am

patphongs:
Thanks for your comments @jenniferthibault ! I've fixed it up according to your specs.

jenniferthibault:
Thanks @patphongs ! 
Style review ✅ 
⛹️ passing to @xtine for technical review

PR #647, opened 2017-03-15
Utility classes for new candidate about page

xtine:
Utility classes to support the new candidate about page. Includes a class for small text, global horizontal rule, and padding for small icon text.

Supports https://github.com/18F/openFEC-web-app/pull/1866

### DIFF
--- a/scss/components/_type-styles.scss
+++ b/scss/components/_type-styles.scss
@@ -197,6 +197,12 @@
   }
 }
 
+.t-small {
+  font-size: u(1.4rem);
+  line-height: u(1.5rem);
+  color: $gray-dark;
+}
+
 .h1 {
   @include heading(h1);
 }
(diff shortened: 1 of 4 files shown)


scss/layout/_layout.scss
```diff
@@ -139,6 +139,18 @@
   padding-right: u(2rem);
 }
 
+.u-small-icon-padding--left {
+  padding-left: u(2.5rem);
+}
+
+// horizontal rule
+
+.hr--light {
```
noahmanger: There's a similar class in `_contact-form.scss` for `.hr--light`. Wanna get rid of that one and also move `.hr--dark` here so they can be together?

scss/components/_type-styles.scss
```diff
@@ -197,6 +197,12 @@
   }
 }
 
+.t-small {
```
noahmanger: 👍 I was just looking for  a utility class for this!

PR #648, opened 2017-03-18
Styling content examples

noahmanger:
This adds styles to support https://github.com/18F/fec-cms/pull/895

### DIFF
--- a/scss/components/_components.scss
+++ b/scss/components/_components.scss
@@ -18,6 +18,7 @@
 @import 'date-grid';
 @import 'documents';
 @import 'dropdowns';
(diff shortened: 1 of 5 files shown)


jenniferthibault:
One small tweak from what I saw in https://github.com/18F/fec-cms/pull/895

In the "Continue learning" section:
- increase margin between heading and thumbnails


image

It should be the same amount of space used in the example thumbnails above:

screen shot 2017-03-20 at 9 34 59 am

noahmanger:
Addressed that change over Slack by using an h3 instead of h4:
image

PR #649, opened 2017-03-20
custom_table_block styles

johnnyporkchops:
## Summary

Added styles to  /scss/elements/_tables.scss for custom table block 
For issue:
https://github.com/18F/fec-cms/issues/813

fec-cms PR: https://github.com/18F/fec-cms/pull/899

scrnsht

I went with vertical and horizontal grid lines because some cells go to two lines, but I can change that

_cc @jenniferthibault 
_cc @xtine


scss/elements/_tables.scss
```diff
@@ -26,3 +26,39 @@ td,
...
+     width: 20%;
+     font-size: 1rem;
+	 font-weight:bold;
+    }
+  } 
+  h2, h3, h4, h5, p{
+    margin:2rem 0 0.5rem 0;
+  }
+  .custom_footnote {
+    margin-bottom:1rem;
```
johnnyporkchops: I will make hound specified corrctions

jenniferthibault:
Thanks @johnnyporkchops ! 

Since visually these should be as close to the  [.simple-table patterns](https://github.com/18F/fec-style/blob/master/scss/components/_table-styles.scss#L1) as possible, I wonder if @xtine could help make code recommendations to get them there? A couple exceptions:
- The numerical data should stay right-aligned (as you have them)
- These tables should remain able to have headings (as you have them)

Other quick changes:
- If there are notes about the table data, can they be styled with the class `t-note`? You can see that style in action on the candidate pages, under the committee cards
- screen shot 2017-03-20 at 11 03 52 am


Changes needed in `.simple-table` to make them more alike
- Add vertical cell borders (to all but outside cells)
- Each cell should have 1rem padding inside it between the cell border and the text



For reference, here's a quick mockup:
screen shot 2017-03-20 at 11 01 13 am

scss/elements/_tables.scss
xtine: A way to avoid these is to either have a CSS linter in your editor active (if you choose to use Sublime Linter or Atom for example, there are packages that automatically lint), or you can run them through online a linter like this: http://csslint.net/

xtine:
Any rem units should be enclosed in a general unit function, so for example:
`margin: u(1rem 2rem)` or `padding-right: u(2.5rem)`.

Also, any colors should be defined in `_variables.scss`. Use a variable already defined or make a new one and try to follow color naming convention.

johnnyporkchops:
@jenniferthibault,
 I am already using t-note class for the footnotes, it looks like your screenshot is using t-italic, is that what we want?
fec-style/styleguide/section-type-styles.html
_cc @jenniferthibault
_cc @xtine

johnnyporkchops:
Regarding the .simple-table patterns, I had considered this when building but as far as I could find, there is no way  to add that class directly to the table element of the tableblock, (either in the block or the template) which is where it would need to go. So I opted for a wrapper,  @xtine , @noahmanger do you know?

johnnyporkchops:
I used scss lint in sublime text and some of the errors hound flag contradict errors that scss lint flags. I removed tabs from the code but hound still says there are tabs. Does anyone have any experience with this happening.

johnnyporkchops:
**Full width:**
scrnsht

**Iphone 5 (320px wide)**
iphone5_tableblock

xtine:
@johnnyporkchops: If are you are using Sublime Text now, I can share a trick with you. In `Preferences.sublime-settings` file (Preferences > Settings), add these lines:
```
	"translate_tabs_to_spaces": true,
	"trim_trailing_white_space_on_save": true
```

And upon save it will automatically translate any tabs to spaces and trim trailing white space.

jenniferthibault:
@johnnyporkchops @xtine I'm experimenting with a new markup template, let me know if this format is useful or just confusing! Click the image to open it larger in a new tab.

Looking at the desktop version, but these changes & differences are observed on mobile as well. I know that most of these styles have been used on our other tables, so hopefully the changes are more about finding the right code to copy & paste than writing it from scratch.

table-edits

(Ignore that my #'s are aligned left, they should be aligned right like you have them!)

Last—on mobile specifically, in the design issue @noahmanger [recommended](https://github.com/18F/fec-cms/issues/550#issuecomment-276844505) that we might just leave them at full-size and use a horizontal scroll: not beautiful, but better than shrinking and cutting off:

> For mobile, what if we simply set it to scroll horizontally? It's not ideal, but it works. 

What do you think?

@xtine would appreciate if you know the points in the code these apply to that might make it easier for John. If there's anything else I can do to clarify please let me know!

xtine:
A few suggestions:
- SCSS files in the `/elements` folder are not for classes, but base styles for HTML DOM elements.
  - I would suggest putting the new custom table class in `/components/_table-styles.scss` as that already has the styles for `.simple-table` which is similar.
- I would suggest to name the new custom table class as `.custom-table` to follow convention like `.simple-table` and `.chart-table`.

scss/elements/_tables.scss
```diff
@@ -26,3 +26,80 @@ td,
...
+  td {
+    border: 1px solid $gray-medium;
+    font-size: u(1rem); 
+    padding: u(0 .5rem);
+    text-align: right;
+  }
+
+  th {
+    border: 1px solid $gray-medium;
+	  font-family: "karla", sans-serif;
```
xtine: This should be `font-family: $sans-serif`

scss/elements/_tables.scss
```diff
@@ -26,3 +26,80 @@ td,
...
+    th {
+      padding: u(0 .3rem);
+      font-size:u(1rem);
+      @include word-wrap(inherit);
+    }
+  }
+}
+
+.custom_table table {
+  @include media(min-width 480px max-width 640px, 12) {
```
xtine: Per Jen's suggestion, perhaps we can remove this size target if we decide it's ok to have horizontal scrolling.

scss/elements/_tables.scss
```diff
@@ -26,3 +26,80 @@ td,
...
+    text-align: right;
+  }
+
+  th {
+    border: 1px solid $gray-medium;
+	  font-family: "karla", sans-serif;
+    font-weight: bold;
+    letter-spacing: -.3px;
+    text-align: center;
+    @include word-wrap(break-word);
```
xtine: We can remove this as the design doesn't have word breaks.

scss/elements/_tables.scss
```diff
@@ -26,3 +26,80 @@ td,
...
+  @include media(min-width 480px max-width 640px, 12) {
+    td {
+      padding: u(0 .5rem);
+      font-size: inherit;
+     }
+
+    th {
+      padding: u(0 .2rem);
+      font-size: u(1rem);
+      @include word-wrap(inherit);
```
xtine: Can remove this, related to line 47.

scss/elements/_tables.scss
```diff
@@ -26,3 +26,80 @@ td,
...
+    padding: u(0 .5rem);
+    text-align: right;
+  }
+
+  th {
+    border: 1px solid $gray-medium;
+	  font-family: "karla", sans-serif;
+    font-weight: bold;
+    letter-spacing: -.3px;
+    text-align: center;
```
xtine: Per design, should be `text-align: left`

johnnyporkchops:
@jeanninehunter, thanks for your detailed comments on the mockup.  Just so you know why I took this down this rabbit hole and could not just use existing CSS. Since this block is available  on page templates with sidebar (like custom page) and on full-width pages (like document page) it can vary between one-third (74.41059%) of .container_main or 100% based on both screen size and parent template. For a table,  this is cumbersome to maintain and get into hacky CSS that breaks from conventions. 
To  address your seven bullet points on the mockup:
1) Added background fill to header to create prominence because the font had to be smaller than usual table headers to avoid overflow/word-breaks
2) I adjusted padding to avoid overflow and word-break as much as possible
3) Yes  the cell border can be $gray-lightest
4) The font is not 1.4 rem at smaller breakpoints to avoid either overflow or word-break as oftern as possible.
5) The text-align-center was one of the things that the Press Office was adamant about although we have been campaigning for them to be willing to change it. Apparently they are quite particular about that.
6) Yes I can take Karla off of the font family, or rather just inherit, my mistake, that was a browser font.
7) Yes the table headers are H4

I think we got the impression that, at least for the Statistical Press Releases, they will frown on the idea of horizontal scrolling, so I thought I had to try to accommodate that, but it is obviously too much of a departure from vernacular. Now I guess at  least we have a proof of concept showing why!

If we go with with horizontal scrolling, I can remove most of this custom styling and just use basic table styles for simple-table.

jenniferthibault:
Thanks for the background @johnnyporkchops — really helpful!

In terms of design sign off, we made sure to get Christian's 👍 from the press office for these designs as part of [the design issue](https://github.com/18F/fec-cms/issues/550#issuecomment-278977894) so I'm comfortable making the changes that bring this in line with the intended styling here. (I think that takes care of # 5. )

Maybe one way to mitigate the template breaks if we moved to horizontal scroll for the tables on small screens is to set the table in a container that contains the scroll. Something like this? @xtine would love to hear what approach you'd take here. (I think this would take care of # 1, 2, 4. )

screen shot 2017-03-23 at 10 35 56 am

jeanninehunter:
Thank you for clarity and context regarding the design, @johnnyporkchops! I appreciate knowing that @jenniferthibault found my feedback helpful during our critique sessions, which I, too, find invaluable for workshopping/troubleshooting project-related challenges.

jenniferthibault:
Oops! Hi @jeanninehunter ! Nice to see you back in fec repos 😸 I think you might have been tagged by accident due to a typo based on autocomplete. You're the first person GitHub autocompletes when you type a name starting with "@-je"

johnnyporkchops:
Changes made for horizontal scrolling. One thing I added was a visual cue (an inset drop shadow ) to indicate that table is scrollable, not just cutoff. *

*This only shows up when the  content is scrollable.  I believe I remember seeing some sort of visual cue like an arrow on scrollable datatable-like content in earlier versions of the site, but I think that was refined and no longer used, but if that is better we can use it or have not visual cue.

mar-28-2017 11-48-41

Looking forward to  @jenniferthibault and @xtine's comments

johnnyporkchops:
It might be easier to pair on this rather than writing out all the caveats, but a couple of examples are:
1)In extending the existing .simple-table styles, some styles had to be explicitly set to ultimately match established table style conventions.
 2)I also hackingly used !imortant because I could not find a way around it, but would welcome input from @xtine on that.

jenniferthibault:
This is looking really nice! I think 👍 from visual style perspective

scss/components/_table-styles.scss
```diff
@@ -253,6 +253,84 @@ h3 + .simple-table {
...
     padding-left: u(4rem);
-    font-style: italic;
+     font-style: italic;
   }
 }
+
+// For custom_table_block in streamfactory
+
+.custom-table {
+  box-shadow: -8px 0px 9px -2px $gray-medium inset;
```
noahmanger: Tiny thing, but if you set this to `-10px 0px 20px -10px #d6d7d9 inset` it will keep the box shadow from coming in on the top and bottom (just from the right).

scss/components/_table-styles.scss
```diff
@@ -253,6 +253,84 @@ h3 + .simple-table {
...
+  display: block;
+  margin-bottom: u(1rem);
+
+  table {
+    @extend .simple-table;
+    width: auto;
+
+    th {
+      @extend .simple-table__header;
+      table-layout: auto;
```
noahmanger: I think this should be on the `table` element, not `th`.

scss/components/_table-styles.scss
```diff
@@ -253,6 +253,84 @@ h3 + .simple-table {
...
+    thead tr th:first-child {
+      border-bottom: 1px solid $primary;
+    }
+  }
+
+  @include media($lg) {
+    box-shadow: none;
+  }
+
+  @include media(min-width 640px max-width 755px 12) {
```
noahmanger: I think we can simplify these and just not worry about hiding it at in-between sizes (just hide it on the large size). What do you think?

scss/components/_table-styles.scss
noahmanger: Also, let's add `line-height: 1;` to tighten up the spacing:
image

xtine:
@johnnyporkchops: Great progressive work on this. Your use of `!important` is fine; because you're extending an existing class you have no choice but to override it this way. The rest of my feedback is line with Noah's.

scss/components/_table-styles.scss
johnnyporkchops: @noah: 
good idea line-height:1

scss/components/_table-styles.scss
johnnyporkchops: Yes, thats a better nuance of the box-shadow settings for this that makes the shadow more subtle and hides the top/bottom shadow better.

scss/components/_table-styles.scss
johnnyporkchops: **"I think we can simplify these and just not worry about hiding it at in-between sizes (just hide it on the large size). What do you think?"**
I agree, I figured that might be a suggestion--which is better because then we are sticking to "mobile-first" and not using any max-width media queries. 
Also those arbitrary breakpoints  I came up with for where to hide/show probably will not be exactly the same across differing environments.

johnnyporkchops:
Regarding Hound errors: I am using scss_lint but just still getting sued to what errors are valid and which to ignore. Two errors that I cannot remove are 
1) !important usage (which @xtine already passed) 
2) use of a selector with a level of applicability of 4 instead of the max, 3.
```tbody th {
      border-bottom: 1px solid $neutral;
    }```
I am using this to remove the dark bottom border that defaults when one chooses the checkbox to make  first column a header as well as the first row in the editing interface. Below are two screenshots of with/without. I can either leave those on or find a way to better target those cells without such a deep selector.
![column_table_border_defaults](https://cloud.githubusercontent.com/assets/5572856/24488582/b97a2cc6-14e5-11e7-88ca-caa9644fd39a.jpg)

File: scss/components/_table-styles.scss [hunk 10]
```diff
@@ -253,6 +253,84 @@ h3 + .simple-table {
...
+    }
+
+    td {
+      @extend .simple-table__cell;
+      border-right: 1px solid $gray-medium;
+      text-align: right;
+      padding: u(1rem) !important;
+    }
+
+    td:last-child,
```
noahmanger: You can get rid of these by unnesting all of these from `table` and just having `.cucstom-table { th { } }` etc. Would that work with the styling?

noahmanger:
Fair point on not being sure which rules to ignore. Both of those are rules that are good to _try_ to work around but sometimes it's just not possible (or is more trouble than it's worth).

One thing you can do (which we should do more of) is [disable the warnings on a case-by-case](https://github.com/brigade/scss-lint#disabling-linters-via-source) basis, like:

```scss
// scss-lint:disable ImportantRule
table {
  border-top: 1px solid #000; !important
}
```

noahmanger:
@johnnyporkchops I just went ahead and unnested the selectors from `table { }` and disabled the `important` error.

johnnyporkchops:
Noah Wrote:" @johnnyporkchops I just went ahead and unnested the selectors from table ..."
Response: Good to know! I did not realize that hound was basing levels of applicability on the nesting structure of the stylesheet, I thought it was basing it on literal nesting levels in the DOM. This is good to know to avoid that error in the future. thanks @noahmanger

PR #653, opened 2017-03-22
New footer updates

xtine:
Styles for new footer changes.
- Adds new `.grid--6-wide`

https://github.com/18F/fec-cms/pull/910

### DIFF
--- a/scss/modules/_footer.scss
+++ b/scss/modules/_footer.scss
@@ -21,31 +21,49 @@
   background-color: $navy;
   color: $inverse;
   font-family: $sans-serif;
-  padding: u(1rem 0);
+  padding: u(2rem 0 1rem 0);
 
   a {
-    border-bottom-color: $inverse;
+    border-bottom: none;
     color: $inverse;
+
+    &:hover {
+      border-bottom: 1px dotted $inverse;
+    }
   }
 
-  .footer-links__column li {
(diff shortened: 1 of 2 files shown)


scss/modules/_footer.scss
```diff
@@ -21,31 +21,49 @@
...
+    &:hover {
+      border-bottom: 1px dotted $inverse;
+    }
   }
 
-  .footer-links__column li {
-    font-size: u(1.8rem);
+  li {
+    font-size: u(1.6rem);
+    padding: (.5rem 0);
```
jenniferthibault: Tiniest adjustment: Could we reduce the line height here so that when an item has to wrap onto the next line it's easier to tell which is a new list item and which is a wrapped line? (As seen on [Tablet](https://cloud.githubusercontent.com/assets/24054/24224810/e4098432-0f22-11e7-8a09-19cbd23fc479.png) from https://github.com/18F/fec-cms/pull/910 )

Here's what I mean using "Plain language" as an example:
screen shot 2017-03-22 at 10 02 14 pm

xtine: Good idea. How does this look to you?
screen shot 2017-03-22 at 10 05 19 pm

jenniferthibault: Much better, thank you!

PR #662, opened 2017-03-30
Fix issue related to jQuery version bump

ccostino:
The update to jQuery caused a regression to appear with one of our selectors, though the selector itself was not set up quite right.  This changeset fixes the issue by adding quotes around the value for the `a[ref="<value>"]` selector.  Please see the comments section of the version 1.12 and 2.2 release notes for details: http://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/

### DIFF
--- a/js/toc.js
+++ b/js/toc.js
@@ -36,7 +36,7 @@ TOC.prototype.addWatchers = function() {
   return this.sections.map(function(idx, section) {
     var elm = document.querySelector(section);
     var watcher = scrollMonitor.create(elm, {top: self.offset});
-    watcher.$menuItem = self.$menu.find('a[href=' + section + ']');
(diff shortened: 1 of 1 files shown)


anthonygarvan:
Also do we need to bump the version for this to get picked up?

ccostino:
Version bump will happen after this is merged. :-)

As for a systematic check, I'm not sure, would have to look into that more.  It'd be nice if there is a way!

ccostino:
Workflow for version bump is this:

* `master` is updated with any changes/hotfixes/etc. that are needed
* When `master` is in a desired state for a release, developer does a `pull` to make sure all latest changes are local for `master`
* Developer runs `npm version <major|minor|patch>`
* Developer runs `git push --follow-tags`

Taken from README

PR #664, opened 2017-03-30
Removing paragraph spacing in lists and .messages

noahmanger:
Sometimes `<p>` tags get nested inside `<li>`s. When that's the case, they shouldn't have bottom margin (unless they're one of multiple, in which case, only the last one shouldn't have it). 

Also if they're inside `.messages` they shouldn't have it either.

Incorrect:
image

Correct:
image

### DIFF
--- a/scss/components/_legal-search.scss
+++ b/scss/components/_legal-search.scss
@@ -100,18 +100,9 @@
   }
 }
 
-.pending-ao__title a {
-  position: relative;
-
-  &::after {
-    @include u-icon-bg($timer, $base);
-    background-size: 90%;
-    background-position: 0 50%;
-    display: inline-block;
-    content: '';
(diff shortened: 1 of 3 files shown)


jenniferthibault:
👍 ➡️  📬 @xtine for code review.

scss/components/_legal-search.scss
```diff
@@ -100,18 +100,9 @@
...
-    content: '';
-    width: 1em;
-    height: 1em;
-    position: absolute;
-    right: -2.5rem;
-  }
+.pending-ao__title {
+  @include u-icon-bg($timer, $base);
+  display: inline-block;
+  padding-right: 2.5rem;
```
xtine: u()

xtine:
👍

PR #666, opened 2017-04-03
Adding styles for dense tables

noahmanger:
Adds custom styles for the occasion when you need an information dense table with `colgroup` and `rowgroup` headers like:

image


jenniferthibault:
Styles 👍 . @xtine could you review the code?

PR #669, opened 2017-04-07
Site orientation tour JS/CSS

xtine:
This is the underlying JavaScript and stylesheets to support the site orientation tour.
- Uses `intro.js` library for tour
- `js/site-orientation.js` for banner and tour interaction
  - Countdown timer until site flip
  - Collapsable banner, state stored with `localStorage`
- End of site orientation tour will initialize a modal:
screen shot 2017-04-07 at 5 23 03 pm


xtine:
Now mobile ready!
wbvtbo5lbo

js/site-orientation.js
```diff
@@ -0,0 +1,203 @@
...
+};
+
+SiteOrientation.prototype.collapseBanner = function () {
+  this.$bannerToggleSection.hide();
+  this.$bannerLessText.hide();
+  this.$bannerMoreText.show();
+};
+
+SiteOrientation.prototype.setBannerState = function () {
+  if (this.$selector.find('.more').is(':visible')) {
```
noahmanger: What about just storing visibility as a property of `SiteOrientation` rather than looking for specific elements to be visible?

js/site-orientation.js
```diff
@@ -0,0 +1,203 @@
...
+    localStorage.setItem('FEC_BANNER_COLLAPSED', 'true');
+  }
+  else {
+    localStorage.setItem('FEC_BANNER_COLLAPSED', 'false');
+  }
+};
+
+SiteOrientation.prototype.startTour = function () {
+  var self = this;
```
noahmanger: Would it make sense to break `startTour` up into smaller methods? Would be easier to test.

js/site-orientation.js
```diff
@@ -0,0 +1,203 @@
...
+
+  // if the user clicks "start a tour" on a page without tooltips
+  // then it will take them to the homepage and start the tour
+  if (typeof TOUR_PAGE == 'undefined') {
+    window.location.href = CMS_URL + '/?tour=true';
```
noahmanger: Just double-checking: is `CMS_URL` a constant set on both the web app and CMS?

scss/modules/_introjs.scss
```diff
@@ -0,0 +1,501 @@
+// CSS to support Intro.js
```
noahmanger: What about just adding a line explaining that these are vendor styles from introjs?

js/site-orientation.js
xtine: It's declared in many places on webapp, and it looks like footer and nav uses it as well. Before next release we should just double check to make sure it's set on both.

PR #673, opened 2017-04-18
Site search styles

noahmanger:
This implements a few changes for the new site search:
- Adds the ability to show unified candidate and committee typeahead results
- Adds the ability to include links to individual contributions and site searches from the typeahead
- Includes miscellaneous styles for displaying site search results


jenniferthibault:
Paired to look at the styles with Noah today — 👍 on styling, over to @ccostino for technical review

scss/mixins/_type-mixins.scss
```diff
@@ -74,7 +74,7 @@
 }
 
 @mixin t-highlight {
-  background: $gray-light;
+  background-color: rgba($aqua, .15);
```
ccostino: Nice, better to be more specific if only setting the one attribute. :-)

js/typeahead.js
```diff
@@ -118,16 +159,22 @@ function Typeahead(selector, type, url) {
...
+    this.dataset = Object.values(datasets);
+  }
+
+  this.init();
 
   events.on('searchTypeChanged', this.handleChangeEvent.bind(this));
 }
 
-Typeahead.prototype.init = function(type) {
+Typeahead.prototype.init = function() {
```
ccostino: Just noting here that changing the function signature will have an impact on anything calling it if it's not also updated.  In this case it's not an issue since any call still passing a value to this method won't be impacted as the method will just ignore the value, but if it were the opposite case (adding an argument), we'd want to be sure we updated all of the calls accordingly to match.

scss/modules/_site-header.scss
```diff
@@ -247,6 +247,15 @@
   }
 }
 
+.utility-nav__search {
+  padding-right: u(2rem);
+  width: u(30rem);
+
+  .tt-menu {
+    z-index: $z-max !important;
```
ccostino: I'm going to agree with Hound here, is there a reason why this is needed?  Okay if there is, just want to make sure we're being deliberate with it!

js/typeahead.js
noahmanger: Ah yep. This is only called by the constructor itself, so I think we're ok.

PR #679, opened 2017-04-19
Removing "No results" messages on typeahead

noahmanger:
This removes the "No matching candidates/committees" boxes on typeaheads when there's no results. As discussed in Slack, this makes more sense because there are still possibly results if you search and we don't want to throw a seeming error message if it's not actually a dead end. See:

image

### DIFF
--- a/js/typeahead.js
+++ b/js/typeahead.js
@@ -85,7 +85,7 @@ var candidateDataset = {
     header: '<span class="tt-suggestion__header">Select a candidate:</span>',
     pending: '<span class="tt-suggestion__loading">Loading suggestions...</span>',
     notFound: Handlebars.compile(
-      '<span class="tt-suggestion__header tt-suggestion__missing">No candidates found matching "{{query}}"</span>'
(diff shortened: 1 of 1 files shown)


jenniferthibault:
I know this is getting a little picky, but seeing "Loading suggestions" appear twice feels like something is going wrong. 

screen shot 2017-04-19 at 9 39 04 pm

I know it's happening because each one is displaying the status for its own category, but do you think it's possible to display just one "Loading" message for all the categories until results are ready?

noahmanger:
I don't think so, without getting really wonky. What about just changing it to "Loading candidate suggestions..." / "Loading committee suggestions"?

jenniferthibault:
Shucks, I knew that was going to be tough!
Maybe "Loading candidates..." / "Loading committees..." ? Keeping it as short as we can feels helpful for visual cluttering & keeping the amount of info to take in down while things are moving.

emileighoutlaw:
Love this! <3

PR #680, opened 2017-04-20
Adding circle calculator class

noahmanger:
Styles to go along with https://github.com/18F/fec-cms/pull/1004


jenniferthibault:
Mea culpa, I should have uploaded the positive version of the icon, instead of the negative of it inside the circle already:

Replaced (L): Before (R)
screen shot 2017-04-19 at 11 45 27 pm

I replaced that image on this PR, but could you adjust the size based on what you're seeing locally @noahmanger ?

PR #683, opened 2017-04-24
Adding class for aligning post content to edge of icon

noahmanger:
Now, when a post title is long and breaks onto multiple lines, everything stays nice and aligned with the icon:

image

cc @jenniferthibault

### DIFF
--- a/scss/components/_posts.scss
+++ b/scss/components/_posts.scss
@@ -64,6 +64,15 @@
   margin-left: u(3rem);
 }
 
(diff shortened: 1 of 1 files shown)


jenniferthibault:
😻 thanks for the detail @noahmanger !

Looking at it all perfectly lined up in context makes me notice how the page titles in the "Other pages" section look like they have a ~1 space indent. Is it possible to line those up as well?

noahmanger:
You're right! Fixed.

PR #688, opened 2017-04-27
Setting href to void

noahmanger:
This helps fix the broken mega menu links in Safari by setting the `href` of the links in the nav to `javascript:void(0)`.


xtine:
Need to update test for:

> initMegaMenu() should remove hrefs from links that have submenus FAILED
> expected '#0' to equal ''

noahmanger:
Ah thanks. Fixed.

PR #692, opened 2017-05-02
Updated loading gif

jenniferthibault:
## Summary

- Replaces loading gif with cleaner version for https://github.com/18F/openFEC-web-app/issues/1977#issuecomment-298738812
- Removes discovered unused image

————————————————————
I changed the size of the image slightly, and tinkered with a lot of export settings, so it would help to see this run locally to make sure it didn't mess anything up. 

cc @xtine @noahmanger


noahmanger:
This works fine, but the stars still look a little jaggy:

image

jenniferthibault:
:shakes-fist: I was afraid of that. I know the cause is that the gif has a transparent background—when you can put it on a solid background you get much crisper edges even at a much smaller file size.

Hold on merging this and I'll see if I can find someone who knows more about exporting gifs with transparency.

jenniferthibault:
Since we weren't able to improve this, should we close it out @noahmanger ?

noahmanger:
👍

PR #695, opened 2017-05-05
Line borders for flat lists

xtine:
New class for line border separators for flat (inline) lists.
In support of: https://github.com/18F/fec-cms/pull/960

### DIFF
--- a/scss/components/_list-styles.scss
+++ b/scss/components/_list-styles.scss
@@ -78,6 +78,23 @@
   }
 }
 
+.list--flat-bordered {
+  @extend .list--flat;
+  padding-bottom: u(.5rem);
+
+  li {
+    &::after {
+      content: '|';
+      padding-left: u(1rem);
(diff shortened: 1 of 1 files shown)


scss/components/_list-styles.scss
```diff
@@ -78,6 +78,23 @@
   }
 }
 
+.list--flat-bordered {
+  @extend .list--flat;
+  padding-bottom: u(.5rem);
+
+  li {
+    &::after {
```
noahmanger: Is there a reason not to just set `border-right` instead of inserting content?

xtine: I can't show a screenshot right now since my fec-cms/style local is busted, but because of the way the padding/margins are set, borders would be longer than the text height.

noahmanger: Gotcha. Ok this works.

PR #706, opened 2017-05-22
Beta cleanup

noahmanger:
This removes remnants of beta language. It also checks for the `CANONICAL_BASE` to see if it's in production, and if so, uses a new localstorage item so that the new banner will be open for everyone when they first visit.

Requires https://github.com/18F/fec-cms/pull/1075 and https://github.com/18F/openFEC-web-app/pull/2062


js/site-orientation.js
```diff
@@ -9,6 +9,9 @@ var helpers = require('./helpers');
 var uri = window.location.toString();
 var uriQuery = helpers.sanitizeQueryParams(URI.parseQuery(window.location.search));
 
+// Use a fresh localstorage item once we're in production
+var STORAGE_ITEM = window.CANONICAL_BASE === 'https://fec.gov' ? 'FEC_BANNER_COLLAPSED_PROD' : 'FEC_BANNER_COLLAPSED';
```
xtine: Can you break this into another line? For readability I'm hesitant on having more than one setter in one line.

js/site-orientation.js
```diff
@@ -9,6 +9,14 @@ var helpers = require('./helpers');
 var uri = window.location.toString();
 var uriQuery = helpers.sanitizeQueryParams(URI.parseQuery(window.location.search));
 
+// Use a fresh localstorage item once we're in production
+var STORAGE_ITEM;
+if (window.CANONICAL_BASE === 'https://fec.gov') {
```
ccostino: No need for `www` here, or it's okay to have the `https://` here as is?

noahmanger: This is just about what we want to set the canonical URL as. Are we setting it to www.fec.gov and redirecting from fec.gov to that? Or vice-versa?

ccostino:
@noahmanger addressed the changes in @xtine's review and I've looked over to check, so I am approving this PR to merge (@xtine is out of the office today).

PR #713, opened 2017-05-26
Data message

noahmanger:
Adding styles for this pattern:
image

### DIFF
--- a/scss/components/_messages.scss
+++ b/scss/components/_messages.scss
@@ -108,6 +108,16 @@
   }
 }
 
+.message--mini {
+  background-position: u(1rem 1rem);
+  font-size: u(1.2rem);
+  padding: u(.5rem .5rem .5rem 4rem);
+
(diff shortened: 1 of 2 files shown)


jenniferthibault:
@noahmanger two small ideas I didn't spot on screenshare. 
What do you think about 
- making the message box the same height as the button? 
- adding some right-hand padding between the end of the statement and the right-hand edge of the box (probably .5-1rem, by eyeballing it)?

Rough sketch of what I mean:

screen shot 2017-05-25 at 9 50 20 pm

I don't think I could see the borders super clearly on screenshare, so sorry for the late ideas!

jenniferthibault:
Small bump to @noahmanger for the comments above, I know this is needed for https://github.com/18F/openFEC-web-app/pull/2083 & the release

noahmanger:
Here ya go:
image

Issue #717, opened 2017-05-30
So that it stands out from the breadcrumbs bar, change "Help for candidates and committees" tab color to `.deep-red`


jenniferthibault:
So that you can visually distinguish the "you are here" state on the tab from the breadcrumbs bar, the color used should be the complement to the primary that's a shade lighter.

On _Data_ & _Legal_ we have and use the primary, `.federal-blue` and it's lighter complement, `.navy`
<img width="1274" alt="screen shot 2017-05-30 at 10 40 52 am" src="https://cloud.githubusercontent.com/assets/11636908/26588870/853da306-4524-11e7-97fb-1935a9bc3535.png">

On _Help for candidates and committees_ we should use the corresponding pattern in red:
`.crimson` and `.deep-red`. Currently, they are both set to `.crimson`.
<img width="1270" alt="screen shot 2017-05-30 at 10 42 34 am" src="https://cloud.githubusercontent.com/assets/11636908/26588956/bbaa1bf4-4524-11e7-9ceb-7bbf732bf6de.png">

Important to note that this is just for the active tab, to show what section of the site you are in when you are a few pages deep. Not for the hover, or focus state.

jenniferthibault:
Closed by #719

PR #719, opened 2017-05-30
Use deep red to indicate parent nav item

noahmanger:
image

Resolves https://github.com/18F/fec-style/issues/717

### DIFF
--- a/scss/modules/_nav.scss
+++ b/scss/modules/_nav.scss
@@ -238,7 +238,7 @@
     padding: u(1rem 3rem 0 2rem);
 
     &.is-parent .site-nav__link__title {
-      border-color: $navy;
+      border-bottom-color: $navy;
     }
 
     &:hover,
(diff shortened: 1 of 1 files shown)


scss/modules/_nav.scss
```diff
@@ -204,7 +204,7 @@
     .site-nav__link {
 
       &.is-parent .site-nav__link__title {
-        border-bottom-color: $secondary;
+        border-bottom-color: $deep-red;
```
jenniferthibault: @noahmanger thank you for hopping on this so fast. One question, I noticed that below on [line 241 ](https://github.com/18F/fec-style/pull/719/files#diff-a7b71e5fa78e81ccd516cf3142ae146eR241) for the corresponding sections in blue, it uses `border-color` instead of `border-bottom-color`. Should they match?

jenniferthibault:
Just noticed that it took the CodeCov thing 3 hours to complete, is something up with that?

scss/modules/_nav.scss
noahmanger: In this case it doesn't make a difference, but good to be consistent. Just updated.

PR #721, opened 2017-05-31
Remove old style guide

xtine:
* Removing references to KSS and older style guide.
* Update README to include link to Pattern Library repo.


ccostino:
Awesome to see! :-)  One question that comes to mind though is do we also need to update the Travis setup in `.travis.yml`?  I remember there was some finagling with getting the 18F-pages stuff setup, seems like we should at least remove that if not incorporate a new setup for the new guide if possible?

xtine:
@ccostino: I removed the references to `deploy.sh` that seemed to be triggering the 18f-pages deployment. Is removing `deploy.sh` and `deploy` on `.travis.yml` enough?

ccostino:
@xtine Yes, I believe that is enough - are we hosting a site for ourselves a-la the former pages-based style guide though?  If so, is anything needed for that?

If not then I think we're good. :-)

xtine:
@ccostino: the new pattern library is hosted on cloud.gov on a different repo, so no setup needed here anymore.

xtine:
@noahmanger: What is https://github.com/18F/fec-style/blob/master/fec-template/public/init.js for? I notice `build-js` runs `browserify -t hbsfy --debug fec-template/public/init.js > fec-template/public/main.js` so I'm wondering what building js on `fec-style` is for.

noahmanger:
That was building the JS for use by the styleguide. Good catch: that should all be ripped out too.

noahmanger:
Looks like we could also remove the `build-sass` and `watch-sass` commands since there's no need to compile CSS here, right? And then I think we can also clear out some of the devDependencies like `hbsfy`, `handlebars`, `less`, `node-sass`, and maybe others?

xtine:
@noahmanger: Great point, the only things I guess that we would need to run are icon compilation and tests?

package.json
```diff
@@ -40,13 +36,10 @@
     "fs": "0.0.2",
     "gulp": "3.8.11",
     "gulp-consolidate": "^0.2.0",
-    "gulp-css-base64": "^1.3.4",
-    "gulp-css-urlencode-inline-svgs": "0.0.5",
```
noahmanger: Looks like we still need this.

PR #732, opened 2017-06-12
Adding styles for nested lists

noahmanger:
Goes along with https://github.com/18F/fec-cms/pull/1105 for styling indented lists.

image

The only thing we're not able to do is to italicize the lower roman lists because you don't have that fine-grained control over the generated value of a list in HTML/CSS.


jenniferthibault:
Bummer about the italicizing, but this looks great and I'm excited for this functionality for the content managers!

Issue #729, opened 2017-06-08
To make content legible, add column margin so that text doesn't collide


jenniferthibault:
This happens on the [_Help for candidate and committees_ landing page](https://www.fec.gov/help-candidates-and-committees/#political-party-committees) 

<img width="1071" alt="screen shot 2017-06-08 at 9 02 35 am" src="https://user-images.githubusercontent.com/11636908/26929968-b4da6764-4c29-11e7-9678-3f2ba1008051.png">

and in table cells on [CMS pages](https://beta.fec.gov/help-candidates-and-committees/filing-reports/post-general-election-reports/) , (as reported by Jamie in https://github.com/18F/FEC/issues/4026)

<img width="657" alt="screen shot 2017-06-08 at 9 05 21 am" src="https://user-images.githubusercontent.com/11636908/26929970-b99deadc-4c29-11e7-9dfa-f140b7a9e0cd.png">

(I know these are caused by two separate things, but the pattern of colliding text is becoming more apparent, so I'm filing it as one.)

jenniferthibault:
The PAC content made an even uglier scenario of this. It looks like the two columns don't have a gutter between them at any screen size

![screen shot 2017-06-16 at 9 38 12 am](https://user-images.githubusercontent.com/11636908/27228917-9125675c-5277-11e7-8f22-ea55ab2ae507.png)

noahmanger:
I have a fix for this, but making the columns narrow yields other problems:
![image](https://user-images.githubusercontent.com/1696495/27247065-75db55fa-52aa-11e7-9373-516432fac5e0.png)

noahmanger:
Nevermind, CSS sorcery FTW:
![image](https://user-images.githubusercontent.com/1696495/27247227-94feb386-52ab-11e7-9e49-077e2cb1b49e.png)

jenniferthibault:
✨ 🔮 ✨ css sorcery! Thank you so much for getting that fix, once the new sections were published it started becoming more obvious.

There would be a separate fix for the table columns later, right? Or are they solved by the same fix? (Serves me right for putting both in the same issue)

noahmanger:
That's a totally separate thing but I just added it to the PR:

![image](https://user-images.githubusercontent.com/1696495/27247734-4bb3a4e8-52b0-11e7-8679-3a7689b1cbe5.png)

PR #737, opened 2017-06-16
Control width of page-link blocks

noahmanger:
This sets a gutter on the page-link blocks used on the Help landing page:
image

It also generates the arrow as CSS pseuod-content so that the line break doesn't happen before the arrow.

Resolves https://github.com/18F/fec-style/issues/729


jenniferthibault:
Yessss this looks awesome. Thanks for the bonus of getting the line break restraint in.

PR #743, opened 2017-07-13
SCSS splitting, images and fonts

noahmanger:
This makes a couple changes to our CSS architecture in order to bring down the size of our CSS files and improve performance.

## Fonts
First, it replaces the previous data-encoded fonts with regular calls to font assets. This brings down the overall size of the stylesheet, which blocks the page render until it's completely loaded.

## Images
Second, it converts the home page hero image and commissioner headshots to jpgs, which cuts down on size considerably.

It also replaces the PNG we use for a header image when printing pages with a background-image on a div, so that the image won't be loaded unless someone's actually printing the page. 

This saves ~334kb on the home page. 

## CSS splitting
Third, it re-organizes the files so that we can be more judicious with what CSS is included on what pages. For example, there's no reason to include data table styles on a CMS article page, and there's no reason to include calendar styles or chart styles on a data table page. 

I started by moving all of the components that were in the `modules` directory into the `components` directory, which mirrors our IA of the pattern library.

Then, I created three top-level .scss files that import everything else:
- `_base.scss`: This imports all of the variables, mixins and libraries necessary for everything else to work. This file typically needs to be `imported` into any new .scss file that will be compiled to a .css file
- `_global.scss`: This imports all of the styles for global components that exist on every page: navigation, footer, page layout, etc.
- `_common.scss`: This imports styles for common page elements, but that aren't necessarily on every page. For example, it includes things like sidebars, heros, etc.

Last, on the web app and CMS sides, I'm submitting a PR that will:
- Create new `global.scss` files, which just import `_global.scss` in order to be compiled to `global.css`. This will be included on every page.
- Create separate page stylesheets that only import the styles necessary. e.g. the data landing page only imports the styles it needs, data tables only the ones they need, etc.

Between the font change and this work, we get significantly smaller CSS files, such as (all numbers gzipped):
- Home page: 410kb -> 152kb (258kb savings)
- Data landing page: 410kb -> 89kb (321kb savings)
- Data table page: 410kb -> 95kb (315kb savings)

With a CSS refactor like this, we'll wanna be super careful and look through everything.


noahmanger:
@xtine I reviewed your changes and they look good. I pushed a couple small things myself. If all looks good, go ahead and approve, merge and publish to NPM as a major version change and then we can bump the version on the app branches.

PR #746, opened 2017-07-19
responsive styles for dense-table-container class

johnnyporkchops:
To solve the responsive issues that arise when using the Contribution Limits Snippet (class="dense-table") in pages with left sidebar and related content on the right. (https://github.com/18F/fec-cms/issues/1093)

Horizontal scrolling and a drop-shadow for a visual cue are used whenever this table is included inside a a div of class="option__content" and also when on a small screen regardless of the container it is in.

Inside class="option__content":
cont_limits_scrnsht_small
 
Inside any other container:
cont_limits_scrnsht_large

## Screenshots

- _Update the styleguide documentation, include a screenshot if applicable._
- _Include a screenshot of the new/updated styles in context (“in the wild”)._

_cc @username_

### DIFF
--- a/scss/components/_table-styles.scss
+++ b/scss/components/_table-styles.scss
@@ -316,10 +316,24 @@ h3 + .simple-table {
 
 // Dense table
 // For tables with a lot of data that need to be condensed
+.dense-table-container {
+  box-shadow: -10px 0 20px -10px $gray-medium inset;
+  overflow-x: scroll;
+
+  @include media($lg) {
+    box-shadow: none;
+  }
+}
+
+.option__content .dense-table-container {
+  box-shadow: -10px 0 20px -10px $gray-medium inset;
+  overflow-x: scroll;
+}
+
 .dense-table {
-  table-layout: auto;
-  border-top: none;
   border-bottom-width: 0;
+  border-top: none !important;
+  table-layout: auto;
 
   td,
   th {
@@ -334,6 +348,7 @@ h3 + .simple-table {
 
   thead th {
     border-bottom: 1px solid $primary;
+    padding: u(.5rem);
   }
 
   tbody th {
--- a/scss/components/_table-styles.scss
+++ b/scss/components/_table-styles.scss
@@ -251,7 +251,7 @@ h3 + .simple-table {
   }
 }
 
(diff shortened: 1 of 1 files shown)


jenniferthibault:
Styles look 👍  to me. Ready for front end review

scss/components/_table-styles.scss
```diff
@@ -316,10 +316,24 @@ h3 + .simple-table {
...
+.option__content .dense-table-container {
+  box-shadow: -10px 0 20px -10px $gray-medium inset;
+  overflow-x: scroll;
+}
+
 .dense-table {
-  table-layout: auto;
-  border-top: none;
   border-bottom-width: 0;
+  border-top: none !important;
```
johnnyporkchops: Added this above the style block with the !important rule to disable to SCSS lint and hound from flagging it:
```// scss-lint:disable ImportantRule```

File: scss/components/_table-styles.scss [hunk 2]
```diff
@@ -316,10 +316,25 @@ h3 + .simple-table {
...
+  box-shadow: -10px 0 20px -10px $gray-medium inset;
+  overflow-x: scroll;
+
+  @include media($lg) {
+    box-shadow: none;
+  }
+}
+
+.option__content .dense-table-container {
+  box-shadow: -10px 0 20px -10px $gray-medium inset;
```
noahmanger: Why is the `.option__content` declaration necessary? Do the styles above not apply?

johnnyporkchops: Good question, I wanted to doublecheck this with you...So when the table is inside 
"option__content", it is in between left and right sidebars (sticky-nav and aside, respectively) so it gets x-scrolling and shadow visual cue. Otherwise when in any larger container, like "main__content--right", it only gets x-scrolling and shadow when the screen itself is small. 
I dont know how this works exactly but when you add an aside to a section, it changes the wraps the original section in"option__content", which is perfect for targeting just this situation. Does that make sense? We can pair on this is if it is confusing.
I also wanted to ask if you think there are other places than resource block that this needs to be in.

noahmanger: Oh! I just didn't understand that at first. That makes sense.

PR #747, opened 2017-07-24
[WIP] Working on six year time selection

noahmanger:
Starting initial work on validating time periods for six year ranges instead of two year. 

Goes with https://github.com/18F/openFEC-web-app/pull/2213


noahmanger:
This is replaced by https://github.com/18F/fec-style/pull/759

PR #753, opened 2017-08-03
Election tweeks

noahmanger:
Styles to support https://github.com/18F/openFEC-web-app/pull/2238 and https://github.com/18F/openFEC-web-app/pull/2231


scss/components/_buttons.scss
```diff
@@ -318,6 +318,8 @@
 // toggle two-candidate view on map comparison
 .button--two-candidates {
   @include u-icon-button($two-candidates, left);
+  background-size: 3rem;
```
xtine: `u()`

PR #754, opened 2017-08-04
Data landing styles

noahmanger:
Various style adjustments and icons to support https://github.com/18F/openFEC-web-app/pull/2243


scss/components/_maps.scss
```diff
@@ -23,6 +23,10 @@
   }
 }
 
+.election-map--small {
+  height: 24rem;
```
xtine: `u()`

PR #759, opened 2017-08-11
Adding a generic date filter for validating periods of time

noahmanger:
This adds a new `ValidateDateFilter` class for use in https://github.com/18F/openFEC-web-app/pull/2213

It works by reading a `data-duration` attribute set on the element, and then as dates are entered, makes sure that they're within a 6 year window. 

demo

~I believe the logic is _not_ that the dates are within 6 years of each other but rather that the *years* of the dates are within six years. Is that right, @LindsayYoung ?~

cc @jenniferthibault and @jameshupp for help sorting out microcopy and interaction details. 
 
Update: per discussion with @LindsayYoung , the validation logic is based on the two dates being within 6 years of each other. I changed the functionality to validate using MomentJS's built in `diff` method.

### DIFF
--- a/js/filter-set.js
+++ b/js/filter-set.js
@@ -14,6 +14,7 @@ var DateFilter = require('./date-filter').DateFilter;
 var ElectionFilter = require('./election-filter').ElectionFilter;
 var ToggleFilter = require('./toggle-filter').ToggleFilter;
 var RangeFilter = require('./range-filter').RangeFilter;
+var ValidateDateFilter = require('./validate-date-filter').ValidateDateFilter;
 
 function FilterSet(elm) {
   this.$body = $(elm);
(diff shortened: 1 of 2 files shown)


jenniferthibault:
What do you think of using the filter validation styling for the error message instead of the alert pattern? If we can do that, it would keep cut down on space and mixing styles here.

Here's two examples of how it's in use now

screen shot 2017-07-20 at 11 00 10 am
screen shot 2017-05-04 at 9 42 11 am

noahmanger:
@jenniferthibault like so? 

image

jenniferthibault:
Yup, thanks!

noahmanger:
Pinging @jameshupp for your thoughts on microcopy.

jameshupp:
Can we change `enter` to `choose`? Better captures the range of ways someone can set the dates. Otherwise I think the copy on that latter example is perfect.

jenniferthibault:
Checking in on this—one microcopy change needed, then good to merge, right?

PR #761, opened 2017-08-16
Adding styles for keyword modals

noahmanger:
Adds styles for the boolean search button as well as the columns in the modal.


scss/components/_buttons.scss
```diff
@@ -510,3 +510,22 @@
   background-image: url('../img/loading-ellipsis-gray.gif');
   background-size: 75%;
 }
+
+.button--keywords {
+  border-bottom: 1px dotted $base;
+  font-size: u(1.4rem);
+  margin-left: 1.5em;
```
xtine: rems?

